PR #82, opened 2016-02-03
Dynamic homepage, Template Option, Tests for Article model

sanajaved7:


home/models.py
```diff
@@ -19,6 +20,59 @@
 class HomePage(Page):
     parent_page_types = ['home.HomePage',]
 
+    TEMPLATE_OPTIONS = (
+        ('Four Story Template', 'Four Story Template'),
```
sanajaved7: Use int instead of string

home/models.py
```diff
@@ -19,6 +20,59 @@
...
+        ('One Story Template', 'One Story Template'),
+    )
+    template = models.CharField(choices=TEMPLATE_OPTIONS, max_length=50, blank=True, null=True)
+
+    promote_panels = Page.promote_panels + [
+        FieldPanel('template'),
+    ]
+
+
+    def get_template(self, request):
```
sanajaved7: Use ints here as well

home/models.py
```diff
@@ -19,6 +20,59 @@
...
+        elif self.template == 'Two Story Template':
+            return 'home/two_story_template.html'
+        elif self.template == 'One Story Template':
+            return 'home/one_story_template.html'
+
+
+    def get_context(self, request):
+        context = super(HomePage, self).get_context(request)
+
+        stories = Post.objects.filter(
```
sanajaved7: sort by date

PR #87, opened 2016-02-06
added map block and template

sanajaved7:
@kjacks - I've moved your work out of master and into a separate branch called data-viz. 

To start working from this branch:
```bash 
git fetch 
git checkout -b data-viz origin/data-viz
```
When your'e ready to push up changes, just use
```bash 
git push origin data-viz
```

### DIFF
--- a/home/blocks.py
+++ b/home/blocks.py
@@ -1,0 +1,24 @@
+from django.db import models
+
+from wagtail.wagtailcore.models import Page
+from wagtail.wagtailcore.fields import StreamField
+from wagtail.wagtailcore import blocks
+from wagtail.wagtaildocs.blocks import DocumentChooserBlock
+
+from wagtail.wagtaildocs.models import Document
+from wagtail.wagtailimages.blocks import ImageChooserBlock
+from wagtail.wagtailimages.edit_handlers import ImageChooserPanel
+from wagtail.wagtailembeds.blocks import EmbedBlock
+from wagtail.wagtailadmin.edit_handlers import FieldPanel, FieldRowPanel,MultiFieldPanel, \
+    InlinePanel, PageChooserPanel, StreamFieldPanel
+
+
+class MapBlock(blocks.StructBlock):
+    data_file = DocumentChooserBlock()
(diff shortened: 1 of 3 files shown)


home/models.py
```diff
@@ -18,6 +18,7 @@
 import django.db.models.options as options
 options.DEFAULT_NAMES = options.DEFAULT_NAMES + ('description',)
 
+from blocks import *
```
sanajaved7: @kjacks - Heads up, we usually avoid "import *" in Python because it can pollute the global namespace. Also you're missing a "." in front of "blocks" which is throwing an error. 

I'd change it to:
```python 
from .blocks import MapBlock
```

home/blocks.py
```diff
@@ -0,0 +1,24 @@
...
+from wagtail.wagtailcore import blocks
+from wagtail.wagtaildocs.blocks import DocumentChooserBlock
+
+from wagtail.wagtaildocs.models import Document
+from wagtail.wagtailimages.blocks import ImageChooserBlock
+from wagtail.wagtailimages.edit_handlers import ImageChooserPanel
+from wagtail.wagtailembeds.blocks import EmbedBlock
+from wagtail.wagtailadmin.edit_handlers import FieldPanel, FieldRowPanel,MultiFieldPanel, \
+    InlinePanel, PageChooserPanel, StreamFieldPanel
+
```
sanajaved7: You might want to review which of these imports you're actually using and only include those. Flake8 is a helpful library you can download to help with that and other PEP8 standards for Python. 

You can install it with:
```bash
pip install flake8
```

You can run it with 
```bash
flake8 file_path
```

sanajaved7:
Hey @kjacks this is really exciting! :+1: :+1: :+1: It's great to see the data viz feature come together. I've left a few comments with suggestions and moving forward, it'd be great if you could also review my pull requests. 

Also, I noticed that there wasn't a migration file to reflect changes to the model so just generate that with the command below and push it up.  
```python
python manage.py makemigrations
```

PR #96, opened 2016-02-12
Dataviz dev

kjacks:


kjacks:
@sanajaved7 I just wanted to check that merging this branch into master wouldn't cause any difficulties

The important files that it touches are urls.py and models.py - each of which with just a couple added lines.

The only other thing that I wasn't exactly sure what to do with are the migration files - will these cause any problems?  They can't just be deleted right?

sanajaved7:
Hey @kjacks I just saw the Travis build failed for this branch. I can check the details of why but do you think it's the models/migration files? If so, I can rebase the branch sometime this weekend and take those out.

sanajaved7:
Oh! Also it looks like you still need to fix the error with importing blocks. It should be: 
```python 
from .blocks import MapBlock
```

kjacks:
Hi Sana,
I changed the import line - so I hope that works better.  It's weird that it was working with the "from blocks import * " too, but hey whichever works!

As for the Travis test, I'm a bit miffed about that because the only thing I changed from the previous commit to the final one was deleting some unused vendor js files - so I'm guessing it is something to do with the migrations changing in the meantime?

sanajaved7:
Hey Kirk - both this branch and the master branch had a migration in the home model at 0006 and this was causing the conflict. To fix this, I pulled your branch, deleted your migrations, rebased from the master branch so that your code was sitting on top of the most recent code, and then created the migrations for this branch which are at 0007 in the home model. 

Also "from block import *" works for some Python versions so that could be causing the error. Also, its discouraged to use import * because it can mess with the global namespace. 

To pull down these changes, use: 
```bash
git fetch 
git reset --hard origin/dataviz_dev
```

Also let's talk more before you merge. I'm having a few issues with running the code. Thanks!

PR #102, opened 2016-02-18
Sidebar basic structure

kjacks:


home/templatetags/top_menu.py
```diff
@@ -23,6 +23,8 @@ def top_menu(context, parent, calling_page=None):
     # programs = Program.objects.in_menu().order_by("title").exclude(location=True)
     programs = []
     location_programs = []
+    print parent
```
sanajaved7: Looks like you left in a couple print statement, I would go ahead and take these out.

sanajaved7:
Hey @kjacks before we merge this into the master branch, I'd like for us to fix some of the markup issues that I ran into last night. We haven't gotten to populating the side bar via database query but this is more about space being left on pages for a sidebar where there shouldn't be - i.e. the home page or org wide pages. I'll shoot you an email about this!

PR #104, opened 2016-02-19
Sidebar basic structure update

sanajaved7:
Hey @kjacks, I like where you're headed with this. On my machine, the sidebar menu was showing up on the homepage so this PR updates some of the logic to generate the sidebar menu only where appropriate. It should be working correctly except for org-wide events, which will require a significant refactor of the Events model. I'll tackle that sometime tomorrow or early next week.

Also just a heads up, I had to change the HTML structure in the `base.html` in order to include the side_bar logic and keep the content block available at all times. The CSS looked a bit funky after those changes so I think we'll need to figure out a solution. Let's touch base tomorrow!

If these changes look good, merge my PR into your branch on Github. Then in the terminal: 
```bash
git checkout sidebar-basic-structure
git pull origin sidebar-basic-structure
```

### DIFF
--- a/home/templates/tags/side_menu.html
+++ b/home/templates/tags/side_menu.html
@@ -1,5 +1,4 @@
 {% load side_menu wagtailcore_tags static %}
-{% get_site_root as site_root %}
 
 <div class="mobile-sidemenu">
   <button class="mobile-sidemenu__toggle">Menu</button>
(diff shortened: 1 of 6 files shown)


kjacks:
Hi Sana, absolutely - what I am doing here is creating basic static structures for each of the elements, not doing any of the dynamic population via the database yet - I figured that would be something we would work together to integrate when you are back in the office.  This looks all great though and ready to go!

PR #107, opened 2016-02-19
Content grid basic structure

kjacks:
I currently built out the design for the content grid within the press release home page, with the idea that we can either use that code/styling to create a more general template or adapt for other content types


press_release/templates/press_release/all_press_releases_home_page.html
```diff
@@ -5,18 +5,65 @@
...
+			</div>
+			<hr>
 		{% endfor %}
-		<br>
-		<br>
-	{% endfor %}
+
+	</div>
+	
+	<!-- {% for release in press_releases %}
```
sanajaved7: Hey Kirk - I'm going to pull out the commented code from the press release template because even though its commented out, it's actually still evaluated by the Django template. To comment things out in the template so that they don't evaluate you can use {# text here #}. I'll remove and merge this branch.

sanajaved7:
Hey Kirk - if you're done with this branch you can go ahead and merge this as well. I removed some of the commented out code from the template.

PR #109, opened 2016-02-24
Sidebar css updates

sanajaved7:


sanajaved7:
Hey @kjacks, I added the wagtail user bar back in since it makes editing the page a lot easier. I noticed it was causing interference with the sidebar menu for mobile screens but that's only an issue when users are logged in to edit content. Let me know if you want to revisit it though and we can talk through it.

kjacks:
Absolutely - I took it outta there because I  wanted to be able to test the toggling of the menu a bit easier (though maybe all the hating on the bird finally got to me...)

PR #110, opened 2016-02-25
Basic post page template

kjacks:
Extended base.html with generic post-page template.  Modified policy paper, press release, and event templates to extend this post-page template.  Styling-wise there are definitely still many more changes that need to be made, but it seemed like a good idea to at least nail down the template inheritance for posts.


kjacks:
Hi @sanajaved7, this is a start as far as how post page template inheritance could work - I would love to talk it through with you though because though it made sense to me as the way post template inheritance  could be structured, I would love to talk it through with you!

kjacks:
Also, a note that the image that is in the post-page template is hard-coded in there (I should have picked a more interesting one...) and would be replaced by the promote image for the post

sanajaved7:
Hey @kjacks I'll take a look at this tonight and send over some feedback!

PR #112, opened 2016-02-29
Search

sanajaved7:


home/management/commands/transfer_data.py
```diff
@@ -0,0 +1,11 @@
+from django.core.management.base import BaseCommand
+from home.management.api.author_transfer_script import run
+
+class Command(BaseCommand):
+	help = """
```
kjacks: Just making sure that the triple quotes are intentional/won't cause any problems

kjacks:
This looks great too!  There is a print statement in there - which is totally fine, just wanted to flag it just in case

sanajaved7:
@kjacks Thanks for catching that! I'm actually changing a few things so will push up those updated this afternoon for you to take a look at.

PR #183, opened 2016-03-08
Simple page

### LINKED ISSUE #170: Separate Simple Page model for Programs
Currently conflicting with sidebar menu logic but might need to be changed in general even if sidebar menu placement in templates is changed.

sanajaved7:
Updated the SimplePage model so that now there is an AbstractSimplePage model which ProgramSimplePage and OrgSimplePage inherit from. 

[Resolves #170]


kjacks:
Looks great!

Issue #192, opened 2016-03-09
Post - Attachments Blocks


rossvanderlinde:
![screen shot 2016-03-09 at 3 33 58 pm](https://cloud.githubusercontent.com/assets/13524901/13649588/66edc034-e60c-11e5-8353-cee390776340.png)

rossvanderlinde:
http://uxrepo.com/static/icon-sets/linecons/svg/attach.svg

PR #212, opened 2016-03-14
Post - Standard Post-Header Basic Templates

kjacks:
- incorporated standard header into press release, podcast, and quoted (for the time being)


sanajaved7:
Good to go!

kjacks:
resolves #192

PR #218, opened 2016-03-15
Fixture edits

sanajaved7:
- Removing the simple page type in the fixture and replacing it with org wide simple page type


kjacks:
@sanajaved7 Feel free to merge this whenever!

PR #222, opened 2016-03-16
Person block

kjacks:
- Updated person and subscribe blocks for the home page and the bottom of post pages
- changed post template inheritance to inherit bottom blocks from abstract template
- modified bylines to conform with first name - last name model change


kjacks:
@sanajaved7 when you have a sec, let me know if this is good to merge!

sanajaved7:
Hey @kjacks what do you think about putting the if statement that checks for the authors in the post_page.html template ```{% if page.post_author.all %} ``` instead of in each of the individual templates?

sanajaved7:
@kjacks That wouldn't affect the name of the field you'd need to check for though, right?

kjacks:
@sanajaved7 defintiely, but then we run into the problem again of the inheritance going the wrong way right?  In that then there needs to be a div in the press release template that populates its contents from the abstract template, instead of the other way around

sanajaved7:
I think we can definitely keep the specific divs in the individual templates like we have them now. I was referring to the the fact that if we don't check for authors and there aren't any authors for a post, we can get something like "None" or some other weird thing when the template renders. I think the check is in most templates already but is missing in some, like blog_post.html for example. If you think it will interfere if its in post_page.html, we just have to make sure it's in each template.

kjacks:
@sanajaved7 The more that is in post_page.html the better, believe me, so if there is a way to do then we definitely should, as long as it can appear in the correct place in the markup.  It is currently in all the posts whose templates have been created (so not blog posts) but I would love to move it if you think it can be done

sanajaved7:
@kjacks Definitely! I think it's worth a try if it can help dry up the code, even if just a little bit. It looks like the if statement could go right above line 21 in post_page.html before we begin iterating over the authors so I'd start there. If you have some capacity to do that, let me know how it goes! Otherwise I can try to tackle it later today.

kjacks:
@sanajaved7 i think I understand the confusion now - line 21 is in the author block at the bottom of the page - which is currently being inherited by all the templates except events.  I thought you were referring to the byline at the top of the page.  I guess what is missing then is just the if statement within post_page.html above line 21

kjacks:
@sanajaved7 It might be worth hopping on a call real quick?

PR #223, opened 2016-03-16
Post byline multiple authors

### LINKED ISSUE #221: Post - Byline for Multiple Authors
- add logic to add appropriate commas, &s depending upon how many authors there are

kjacks:
- incorporates logic for formatting of 1, 2 or 3+ authors in byline
- changes template inheritance to place author and date in post_page.html template, with post-header, byline-prefix, and post-body blocks left in individual post templates

Resolves #221

### DIFF
--- a/book/templates/book/book.html
+++ b/book/templates/book/book.html
@@ -25,6 +25,15 @@
 
 {% endblock %}
 
+{% block byline-prefix %}
+
+	{% if authors|length > 1 %}
+		Authors:
+	{% else %}
+		Author:
+	{% endif %}
+
+{% endblock %}
 
 {% block post-body %}
 
--- a/book/templates/book/book.html
+++ b/book/templates/book/book.html
@@ -5,7 +5,7 @@
 
 {% block body_class %}template-book{% endblock %}
 
-{% block post-content %}
+{% block post-header %}
 	{% image page.story_image original as story_image %}
 
 	<div class="post-header">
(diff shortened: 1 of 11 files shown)


kjacks:
@sanajaved7 let me know what you think! Slash if it's ok to merge

book/templates/book/book.html
```diff
@@ -23,24 +23,22 @@ <h2 class="post-header__text__sub-headline">{{ page.sub_headline }}</h2>
...
-				{% for author in authors %}
-					<a href="{{ author.author.url }}">{{ author.author.first_name }} {{ author.author.last_name }}</a>
-				{% endfor %}
-			</h5>
-			
-		{% endif %}
+{% block byline-prefix %}
 
-	    <h5 class="post-date"> {{ page.date }} </h5>
+	{% if authors|length > 1 %}
```
sanajaved7: Yay! Glad to see you used the template length feature

mysite/templates/post_page.html
```diff
@@ -7,9 +7,36 @@
...
+
+	    {% if page.post_author.all %}
+		    	
+		    <h5 class="post-author">
+		    	{% block byline-prefix %}
+				{% endblock %}
+				
+				{% for author in authors %}
+					{% if forloop.revcounter0 >= 2 %}
+						<p><a href="{{ author.author.url }}">{{ author.author.first_name }} {{ author.author.last_name }}</a>,</p>
```
sanajaved7: Hey @kjacks I'd recommend moving the comma logic into a template tag. We'd have a lot more control and would be able to use Python for a cleaner solution. Also we could reuse elsewhere if necessary. I think [simple template tags](https://docs.djangoproject.com/en/1.9/howto/custom-template-tags/#simple-tags) should do the trick.

kjacks:
@sanajaved7 - woah, didn't know you could make your own template tags - I was looking through and thinking there might be a pre-built list separator kinda thing, but if we can make our own that would definitely be a cleaner solution

PR #236, opened 2016-03-21
Expert & Org-wide Our People pages

kjacks:
- Implemented templates for experts and org-wide our people pages
- added simple sidemenu to our work and about pages


sanajaved7:
Also good to go :+1:

PR #237, opened 2016-03-21
Fix Post Home Page Title Issue

kjacks:
changed inheritance block name within org wide and program post home pages from title to page-title to avoid naming conflict


sanajaved7:
Feel free to merge!

PR #309, opened 2016-03-28
Person model changes 3

### LINKED ISSUE #306: Our People classification
Per Mark's feedback, we don't need a distinction between national and program fellows. So we can just have 'Fellow' listed in the role dropdown

sanajaved7:
- Combines roles for New America Fellow and Program Fellow to just "Fellow"

[Resolves #306]


georgiamoon:
Oop -- late to the game here, but I thought that we weren't combining program fellow and new america fellow? or are we clarifying the designation differently?

sanajaved7:
Hey @georgiamoon - I believe this was the result of a discussion between @rossvanderlinde and Mark. My understanding from @rossvanderlinde is that the distinction between program and NA fellows can be blurred and since we can associate people with programs we didn't need a separate role for program fellow. Is that correct @rossvanderlinde?

rossvanderlinde:
@georgiamoon AMS asked me to ask MS for a decision. I can explain further in person if necessary

georgiamoon:
OK -- just checking that there's still a way to designate, as that designation matters to people in other contexts.

kjacks:
:+1:

PR #319, opened 2016-03-29
Add get page type function

kjacks:
- added function to models for types that appear in content grids to identify their as a class name


sanajaved7:
Also good! :+1:

PR #331, opened 2016-03-30
Post author block fixed height

### LINKED ISSUE #294: Post - Book/Policy Paper Change Story Image to Cover Image
- when model field is changed, update which field post header is pulling from

### LINKED ISSUE #330: Post - Fixed Height for Author Block
- fix issue of overflow where author short bio is too short

kjacks:
- set fixed height for post-author blocks and vertically centered text - Resolves #330 
- removed extraneous image tag in book and policy paper templates - Resolves #294

### DIFF
--- a/book/templates/book/book.html
+++ b/book/templates/book/book.html
@@ -6,16 +6,14 @@
 {% block body_class %}template-book{% endblock %}
 
 {% block post-header %}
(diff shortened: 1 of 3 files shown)


sanajaved7:
:+1: good to go!

PR #381, opened 2016-04-12
Bio/Author Block With No Image

### LINKED ISSUE #378: Bio Page/Author Block - Disappear Circle when No Image

kjacks:
- Added logic to handle author block and bio page header with no person image
- changed byline to use author field from context instead of post_author to ensure correct ordering

Resolves #378


sanajaved7:
Looks good! :+1:

PR #386, opened 2016-04-13
Programs Dropdown Right Column Formatting

### LINKED ISSUE #383: Header - Programs Dropdown Location/Channels Column Width
- set breakpoints to change width of right hand column on smaller screen sizes

kjacks:
- added breakpoints to change programs dropdown right column width to match main column widths
Resolves #383


sanajaved7:
Also good!

PR #402, opened 2016-04-13
Fixing Tests to Provide Required Fields

sanajaved7:
- Django upgrade is enforcing (thankfully) required fields to have default values. Updating the tests to provide these required fields.


kjacks:
Feel free to merge!

PR #424, opened 2016-04-15
Removing the default value from the Event soundcloud URL field

### LINKED ISSUE #406: Event Soundcloud Field Change
- Remove default value

sanajaved7:
[Resolves #406]


kjacks:
👍

PR #427, opened 2016-04-18
Short Bio - Rich Text Filter and Icon Size

### LINKED ISSUE #422: Short Bio - Add rich text filter

kjacks:
- added rich text filter to handle rich text in short bios
- formatted person icon image to cover when image smaller than circle
Resolves #422


sanajaved7:
This looks good to merge 👍 and as a side note, we should probably check what the other RichText fields are and ensure they have the filter in the template as well. I can create a separate issue for that.

PR #469, opened 2016-04-19
Added Datepicker to Content Grids

### LINKED ISSUE #230: Content Grid - Date Picker
Date picker. This is the design currently - let's talk further about it

screen shot 2016-03-21 at 11 41 27 am

kjacks:
- Implemented jQuery UI Daterangepicker 
Resolves #230


event/templates/event/all_events_home_page.html
```diff
@@ -4,6 +4,42 @@
 {% load wagtailimages_tags %}
 {% load utilities %}
 
+{% block extra_scripts %}
```
sanajaved7: Hey @kjacks any way we could import this script from one place so we don't have to duplicate it in both templates?

Issue #494, opened 2016-04-20
Post Author Block - Text Default


kjacks:
- if no person description, default to author name

PR #510, opened 2016-04-21
Post Author Block - Set Fallback if no Short Bio

### LINKED ISSUE #508: If author has no short bio or photo, word "Host" appears with nothing after it.
Under the word "Host:" or "Author:" at the bottom of the page if the author has no bio then nothing shows up.

Best solution is probably to make sure everyone has a bio.

Creating a new 'May Cleanup' tag for things for temp and comms team to work on prior to launch.

kjacks:
Resolves #508 #494


sanajaved7:
👍

PR #542, opened 2016-04-22
Attaching subprogram to program 

### LINKED ISSUE #526: Add logic to attach subprogram to parent program if user does not

sanajaved7:
Logic to connect subprogram to parent program under which its been created if user does not select the program

[Resolves #526]


kjacks:
👍

PR #581, opened 2016-04-27
Adding update index command for elasticsearch to the config file

sanajaved7:
Adding command to automatically updating database index for elasticsearch upon deployment


kjacks:
👍

PR #584, opened 2016-04-27
Add Image Source & Caption to Templates

### LINKED ISSUE #248: Photo credit
screen shot 2016-03-22 at 1 51 33 pm

kjacks:
- added image source to all post header images
- added image source and caption to all post image blocks
- post-launch will look into adding source/captions to images in richtext fields (according to Wagtail docs, no direct way to do) -> would need to use script to append to the DOM
Resolves #248


blog/templates/blog/program_blog_posts_page.html
```diff
@@ -7,12 +7,16 @@
...
 {% block content %}
-	<div class="post-header">
-		<div class="post-header__text">
-		    <h1 class="post-header__text__heading">{{ page.title }}</h1>
+	<div class="post-header-container">
+		<div class="post-header">
+			<div class="post-header__text">
+		    	<h1 class="post-header__text__heading">{{ page.title }}</h1>
+		    	{% if page.subheading %}
+		    		<h2 class="post-header__text__subheading">{{ page.subheading }}</h2>
```
sanajaved7: Hey @kjacks I noticed the richtext filter has been removed from the subheading field - are we handling that in another way?

kjacks: @sanajaved7 I'll put it back in there but within the conditional.  I totally forgot we had made this specific subheading richtext?

kjacks: I was trying to remember why we made that change

sanajaved7: Not sure either - but it's quite weird it's the only subheading that's a richtext field. I guess because it's for the blog landing page and slightly different.

PR #634, opened 2016-04-28
Content Type Line

kjacks:
- created template tag to generate content type line in hero grid, featured work, and search results
- maps from content type to label (ex. Articles and Op-eds -> Article)


sanajaved7:
Hey @kjacks on the technical side I think this is a good way to handle it but I wonder if it will throw users off that for some pages you will see the type and others you won't. Did you and Ross discuss that already? Don't want to repeat the conversation if you have.

kjacks:
@sanajaved7 We did - the decision to not display the type on those pages came from the fact that for  those pages the type name is kind of more for internal purposes and would have less meaning for the user - or just wouldn't really mean anything outside of the editor

sanajaved7:
@kjacks Got it - feel free to merge!

PR #667, opened 2016-05-03
Removed leading zero from date format

### LINKED ISSUE #602: Date Format - remove leading 0
- should be March 9, 2015 not March 09, 2015

kjacks:
Resolves #602


sanajaved7:
👍

Issue #653, opened 2016-04-29
Desktop: [Firefox/IE] Field of calendar with date loses frame after date selection


rossvanderlinde:
Desktop, Browser: Firefox/IE
Page: http://newamerica-cms-dev.us-west-2.elasticbeanstalk.com/policy-papers/

Open page
Select date range 
Confirm
- There is no frame for field now
Click anywhere on page
- Frame appears again

[--- This issue is linked with an Asana task
#attachments [2016-04-28_8-00-44.mp4](https://app.asana.com/app/asana/-/download_asset?asset_id=118079082793648)
#url https://app.asana.com/0/0/118079082793645/f
---[aa]

PR #669, opened 2016-05-03
Content Grid Controls Formatting

kjacks:
- set formatting for controls across breakpoints
- added update search and reset buttons, with disabled states
- Resolved I.E. date input field losing focus #653


sanajaved7:
👍

PR #673, opened 2016-05-03
Set Experts Page to Our People

### LINKED ISSUE #668: Set Experts Page to Our People

kjacks:
- set header links to our people instead of experts page
- removed experts from about sidemenu

Resolves #668

### DIFF
--- a/home/templates/tags/top_menu.html
+++ b/home/templates/tags/top_menu.html
@@ -57,7 +57,7 @@
                 <li class="mobile-menu__link__secondary"><a href="/press-releases/">Press Releases</a></li>
             </ul>
         </li>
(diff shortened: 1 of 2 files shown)


kjacks:
@sanajaved7 We should talk about this briefly.  Ross and I were talking about the experts page and he thought that without the ability to search by expertise area, the experts page was exactly the same as the our people page, so therefore for now we should just have the our people page.  The header experts links would now just go to the our people page

kjacks:
I left the experts model and template in there because it is something that we will hopefully build out soon and won't want to have to crazily change structure, but let's definitely touch base on this

sanajaved7:
@kjacks that sounds like a good plan to me while we figure out how we'd like to build out that page 👍

kjacks:
@sanajaved7 great!

PR #677, opened 2016-05-04
Org wide Search Page 

sanajaved7:


search/templates/search/search.html
```diff
@@ -19,16 +19,20 @@ <h1 class="post-header__text__heading">Search New America</h1>
...
+                <div class="content-grid__controls__filter">
+                    <input type='text' name='query' placeholder='Search' autocomplete="off"></input>
+                    <button class="search-box__submit" type="submit">
+                        {% include 'ui_elements/svg/magnifying-glass.html' %}
+                    </button>
+                </div>
+            </div>
+        </form>
+
         <div class="content-grid__controls__right">
```
kjacks: @sanajaved7 the formatting is breaking because some elements have lost their class names - i.e. the input field should still have the "search-box__input" class, the outer div needs "row" etc.  So if it's possible to keep those classes in the markup that would be great.

kjacks: I can also go back in there and add them back in if need be, but it seems like we shouldl keep them at this point if we can

PR #680, opened 2016-05-04
If statement for images in events template

sanajaved7:
Found a bug where events that were very different had the same image. Turns out there wasn't a if statement checking if that story had an image and so it was calling the image from the previous event in the loop.

Added the if statement and things should be good now.


kjacks:
Nice catch, looks good to go!

PR #686, opened 2016-05-05
Events - Past/Future

### LINKED ISSUE #672: Events
- add events to mobile top menu (done)
- remove past and future from org wide events sidemenu? (done, kept)
- add logic to only display rsvp button when event is future (done)
- set event datepicker options
- military time in editor?
- start/end times in editor?
- 12 results per page
- order results by time in addition to date (done)

kjacks:
- set default event datepicker options
- added past and future events links to mobile menu
- set past and future on org wide events sidemenu
- added logic to only display rsvp button when event is future
- set event past and future predefined datepicker options
- order event results by time in addition to date

Resolves #672


home/templatetags/utilities.py
```diff
@@ -118,3 +120,32 @@ def generate_dateline(post):
 
 	return mark_safe(ret_string)
 
+@register.simple_tag()
```
sanajaved7: Hey @kjacks could you add some documentation to this template tag? It looks pretty complicated and I think it'd be good to jot down it's functionality while it's still fresh in your mind. Thanks!

PR #699, opened 2016-05-05
Database migration script and requirements.txt changes

sanajaved7:
Hey @kjacks heads up - this branch has a number of print statements but I actually need them when tracking the transfer so once we're done with the migration I can take them out of the repo. Thanks!


kjacks:
Looks great!

PR #688, opened 2016-05-05
Attachment Box Styling

kjacks:
Resolves #632

PR #701, opened 2016-05-05
Pagination Results Per Page & Our People Pagination

kjacks:
- changed number of results per page from 10 to 12
- added pagination to program our people and central staff etc people pages
- removed content filters from our people pages


sanajaved7:
Pagination also changed here (and in #688)

kjacks:
@sanajaved7 yeh, it's actually the same commit, but got folded into both branches, but I rebased so it should be all set!

PR #720, opened 2016-05-06
Implemented Ellipsis for Story Excerpts at Character Limit

kjacks:


sanajaved7:
Hey @kjacks I thought we were holding off on this? Also none of the excerpts are going to be greater than 140 characters because the database validation won't allow it and that's why the excerpts are cut off.

kjacks:
@sanajaved7 Ross said to go ahead with it as a temporary fix - we can definitely hop on a call to talk it through - but it ended up being pretty straightforward.  I'm sure there will be a few instances where it ellipsifies something that is rightfully at the character limit, but it seems like at least a solid temporary catch for legacy content?  Ass for the >= instead of the ==, I know that there is no real way for the limit to get above that amount now, but programatically wanted to handle all cases of the conditional just in case

mysite/assets/js/modules/story-excerpt-ellipsis/index.js
```diff
@@ -0,0 +1,21 @@
...
+
+var maxLength = 140;
+
+//
+// catch for legacy content with cut off story excerpts - adds ellipsis to any story excerpt appearing in the content grids or picture grids
+//
+export default function addAllStoryExcerptEllipsis() {
+	function addEllipsis() {
+		var text = $(this).text();
+		if (text.length >= maxLength) {
```
sanajaved7: Hey @kjacks I think I mentioned this before, but I don't think we need to check for text length here because the summaries have been cut off at 140 in order to be added to the database.

kjacks: @sanajaved7 As for the >= instead of the ==, I know that there is no real way for the limit to get above that amount now, but programatically wanted to handle all cases of the conditional just in case. I can definitely change this to just == if you think that would be better, but as a practice I tend to think it's better to handle the full range of the conditional

sanajaved7: Well I'm just wondering if the length check is actually redundant. If there's punctuation at the end (which we're checking for) then that means the summary was adequate otherwise, we add the  ellipses. Thoughts?

PR #743, opened 2016-05-09
Programs Dropdown - Show Subprograms Conditional

### LINKED ISSUE #726: Allow Initiatives to Be Removed from Programs Dropdown
June project: can we limit subprograms that appear in the programs dropdown to only those that have "Show in Menus" selected?

This will allow us to not show retired/archived initiatives while still categorizing content in that initiative.

kjacks:
- added conditional to only display subprograms if show in menus is checked

Resolves #726

### DIFF
--- a/home/templatetags/top_menu.py
+++ b/home/templatetags/top_menu.py
@@ -39,14 +39,18 @@ def top_menu(context, parent, calling_page=None):
     }
 
 
-# Retrieves the children of the top menu items for the drop downs
+# Retrieves the children of the top menu items for the drop downs - will only add iniatives where show_in_menus field is set to true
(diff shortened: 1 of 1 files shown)


sanajaved7:
Hey @kjacks I think there might be a easier way to implement this so if it's non-essential can we hold on this?

kjacks:
@sanajaved7 I've gotten feedback from a couple people at this point about this feature - would be happy to talk through other solutions, or merge this approach and refactor later, whichever makes sense on your end!

sanajaved7:
Hey @kjacks just opened PR #918 with a different approach which I think is a bit simpler but happy to discuss!

PR #754, opened 2016-05-10
Landing Page Carousel Conditional

### LINKED ISSUE #732: Hide both carousels on a program homepage if no events are added to events carousel
Not essential that's it's done pre-launch but it would be good if it's possible

kjacks:
- added conditional to display featured carousel on home page and program landing page when items featured in carousel - Resolves #732


sanajaved7:
👍

PR #757, opened 2016-05-10
Content/Picture Grids - "No Results Found" if none returned

### LINKED ISSUE #730: No Results in Content Grid
If no results are found in a content grid (incl after using a filter), please can it say 'No Results' where the results would otherwise have appeared

kjacks:
Resolves #730


sanajaved7:
Also looks good!

PR #775, opened 2016-05-11
Handling Redirects from current site to future site via Wagtail

### LINKED ISSUE #450: Redirects
Potentially assigned to Nat? Can we add him to this Github repo https://app.asana.com/0/113884670727555/113884670727555 ?

Need to ensure .net still redirects, while forwarding current .org/program/page structure to new .org/program/content-type/page structure.

sanajaved7:
Using wagtail Redirect model, the script creates objects that map old (current) URLs to new paths

[Resolves #450]

### DIFF
--- a/home/management/api/newamerica_api_client.py
+++ b/home/management/api/newamerica_api_client.py
@@ -165,6 +165,19 @@ class NAClient:
 				for post in post_set['results']:
 					yield post, program_id
 
+	def get_posts(self):
+		"""
+		Gets all the content type of Post from the old database API
+		for all programs and creates new objects in the new database 
+		using the Event model
+		"""
+		for program in self.client.get(self.api_url + 'programs').json():
+			program_id = program['id']
+			self.activate_program(program_id)
+			for post_set in self.get_data('posts'):
+				for post in post_set['results']:
+					yield post, program_id
+
 
 	def program_content(self, program_id):
 		"""
--- a/home/management/api/newamerica_api_client.py
+++ b/home/management/api/newamerica_api_client.py
@@ -168,8 +168,7 @@ class NAClient:
 	def get_posts(self):
 		"""
 		Gets all the content type of Post from the old database API
-		for all programs and creates new objects in the new database 
-		using the Event model
+		for all programs
 		"""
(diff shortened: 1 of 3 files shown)


home/management/api/newamerica_api_client.py
```diff
@@ -165,6 +165,19 @@ def get_in_the_news(self):
 				for post in post_set['results']:
 					yield post, program_id
 
+	def get_posts(self):
+		"""
+		Gets all the content type of Post from the old database API
+		for all programs and creates new objects in the new database 
+		using the Event model
```
kjacks: @sanajaved7 is it just the event model, or whatever the post type is?

home/management/api/redirect.py
```diff
@@ -0,0 +1,81 @@
...
+            for program in post['programs']:
+                old_path = '/{0}/{1}'.format(
+                    mapped_programs[str(program)], post_slug
+                )
+                new_redirect, created = Redirect.objects.get_or_create(
+                    old_path=old_path
+                )
+                new_redirect.redirect_page = new_page
+                new_redirect.save()
+                print(old_path)
```
kjacks: @sanajaved7 just flagging this print statement to make sure we want to keep it for debugging, etc.

kjacks:
@sanajaved7 this looks great - are we then handling all redirects through wagtail, or is this just for posts and people?

home/management/api/redirect.py
sanajaved7: @kjacks yep want to keep it there just to make sure script is working + moving along!

sanajaved7:
@kjacks I think this actually allows us to handle all the redirects through Wagtail actually. I can elaborate more during our check-in but the script creates objects in Wagtail's Redirect model using the old and new path for all pages. Pretty cool right?

kjacks:
@sanajaved7 absolutely! That would be great if we could control it all from here - as long as there are no performance implications? I feel like wagtail would have redirects figured out pretty well though

sanajaved7:
@kjacks Definitely worth continuing to test in staging but for what it's worth, it's a pretty smooth redirect locally at least so I'm hoping that's a good sign 😄

PR #778, opened 2016-05-11
Person Conditionals

### LINKED ISSUE #766: Search Results - Person Content Type Display Inconsistency
screen shot 2016-05-10 at 5 31 32 pm

### LINKED ISSUE #770: Person bio tweaks
If the individual does not have any position at New America, it is currently displaying 'None'. Should just be blank. 

Also, for any external/former staff, please do not show their email address, even if the email address is filled in. 1) We can't keep track of these as they change over time, 2) We currently have hundreds of them set up as incorrect name@null.com type addresses which we dont want to display

kjacks:
- Added conditional to eliminate 'None' in position line in bio from content transfer
- added conditional to display contact info on bio page if person has an email and is not former staff/external author
Resolves #770 

- added fallbacks in search results page to show person title if no short bio and 'Person' page type with or without image
Resolves #766

### DIFF
--- a/home/templatetags/utilities.py
+++ b/home/templatetags/utilities.py
@@ -172,3 +172,11 @@ def is_future(date):
 		return 1
 	else:
 		return 0
+
+@register.simple_tag()
+def person_display_contact_info(page):
+	if (page.email):
+		if (page.role != "External Author/Former Staff"):
+			return 1
+	
+	return 0
--- a/mysite/assets/scss/components/_content-entry.scss
+++ b/mysite/assets/scss/components/_content-entry.scss
@@ -86,6 +86,11 @@ $content-entry-text-gutter: 10px;
 			font-size:16px;
 			line-height: 150%;
 		}
+
+		&__person-position {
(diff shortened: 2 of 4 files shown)


kjacks:
@sanajaved7 if you wouldn't mind taking a look at this one when you get the chance - not urgent though

home/templatetags/utilities.py
```diff
@@ -172,3 +172,11 @@ def is_future(date):
 		return 1
 	else:
 		return 0
+
+@register.simple_tag()
```
sanajaved7: Hey @kjacks can we actually handle this logic in the template? I'm not sure myself but let me know!

kjacks: @sanajaved7 I went for a tag because the check needs to happen in multiple places across the site, so figured it made more sense not to clutter the templates with lots of nested if statements, but could work either way

sanajaved7:
@kjacks sounds good! Feel free to merge!

PR #779, opened 2016-05-11
Refactor Frontend Build

kjacks:
- adds eslint for automated linting testing
- folds webpack config dev and production files into one file with parameter setting differences between the two
- removes vendor js files from the head tag and loads them within mysite.js
- exposes external functions from vendor files and user-defined modules, with single document.ready function to call them
- removes vendor css files form the head tag and imports them through mysite.scss
- moves all non npm vendor js and css files from static/ to assets/ folders


kjacks:
@sanajaved7 Let me know if you want to talk through any of these changes.  I will add more extensive documentation to the scripts, and I will talk with Peter about documentation for webpack-specific documentation

sanajaved7:
Hey @kjacks do you have that documentation ready to add here? Would be good to include before we merge.

kjacks:
@sanajaved7 Peter hasn't been on today, so I haven't gotten a response yet from him

PR #798, opened 2016-05-12
Pagination with query

### LINKED ISSUE #709: Pagination - Keep Query Parameters When Switching Pages

sanajaved7:
Updating pagination to pass query parameters when they exist onto following pages on:
- Org wide post homepage
- Program post homeage
- Org wide events homepage
- Program events homepage

[Resolves #709]

@kjacks FYI - I took out the code for pagination at the top right of pages from the templates since that was being hidden anyway but let me know if you need that in there.


kjacks:
@sanajaved7 pagination is back in to the top of the page, feel free to merge

PR #806, opened 2016-05-13
404 and 500 Pages

### LINKED ISSUE #741: Add a 404 page
Something along the lines:

"Oops couldn't find it! Try Searching?"

include search module

kjacks:
- set custom 404 search results template and 500 template
- set 404 handler to use custom view
Resolves #741


search/views.py
```diff
@@ -36,3 +38,21 @@ def search(request):
         'search_query': search_query,
         'search_results': search_results,
     })
+
+
+def search404(request):
+    search_query = os.path.basename(os.path.normpath(request.path))
```
sanajaved7: Hey @kjacks is this so that we can turn the url into the search query? Just want to make sure I understand the difference with using this vs. request.GET.get

kjacks: @sanajaved7 absolutely - essentially the difference between where the 404 search is getting its query versus where the regular search is getting its query is that the regular search is looking for a parameter in the querystring part of the url, while the 404 search is looking for whatever is in the final section of the url

kjacks: this line gets the url and parses out only the final section of it in order to search, ignoring any querystring parameters as well

sanajaved7:
Thanks for clarifying and for adding documentation!

sanajaved7:
Good to merge 👍

PR #849, opened 2016-05-16
Changing elasticsearch config to auto index

### LINKED ISSUE #833: Slow updating reports
From Alison:

I'm having trouble featuring an upcoming event on the OTI homepage. I added the event to OTI events, and it shows up on the future events section of the production site, but it took an hour before it showed up in the events section in Wagtail. It now shows up in the OTI events section of Wagtail, but when I go to feature it on the OTI homepage, either when I search for it or when I navigate to the OTI events section in Wagtail, it doesn't show up. I've tried changing the date, renaming it, and untagging it to OTI, but none of things have seemed to work.

From David Sterman:

I created an article page for 
What the 2016 Presidential Candidates Get Wrong About the Future of War
We would like to have it be a lead story for Future of War's front page, but it doesn't show up  in the backend search to select it. It does however show up under article tagging. Is there a fix for this?

(Note: I searched for the page and it shows up, suggesting there was also a delay in it showing up for David).

sanajaved7:
[Resolves #833]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,9 @@ venv
 *.log
 
 *.csv
+*.crt
+*.key
+*.csr
 
 home/management/api/article_images/
 home/management/api/images/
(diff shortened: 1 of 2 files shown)


kjacks:
👍

PR #853, opened 2016-05-16
Updating queries for drop-down menus 

### LINKED ISSUE #807: Update query that populates drop down menus
- Need it to only include programs that have been selected to show in menus, currently showing all programs, including archived ones

sanajaved7:
Now will only show programs and subprograms that are published and have been marked as "show in menus"

[Resolves #807]


kjacks:
👍

PR #860, opened 2016-05-17
Redirect/404 bug

sanajaved7:
@kjacks - Let's talk more in the morning but this line in the views for the 404s was breaking all the redirects. Took it out and the 404 page is still working.

### DIFF
--- a/mysite/urls.py
+++ b/mysite/urls.py
@@ -18,8 +18,6 @@ urlpatterns = [
     url(r'', include(wagtail_urls)),
 ]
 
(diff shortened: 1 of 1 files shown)


sanajaved7:
@kjacks update - the search results aren't actually populating so we'll either need to fix this or pick one for the time being while we find a solution.

sanajaved7:
@rossvanderlinde going to make a call to fix this so redirects work (esp. for the first few days of the site) and then we should figure out a solution in the meantime.

PR #875, opened 2016-05-17
Text Wrap - Cross Browser

### LINKED ISSUE #845: Need max height or something to deal with lead story overflow
See here https://na-production.herokuapp.com/cybersecurity-initiative/

kjacks:
- updated handle-text-overflow mixin to best handle cross-browser approaches to word-wrapping and hyphenation
- set max-height to handle text overflow for lead stories - Resolves #845


sanajaved7:
Good to merge!

PR #876, opened 2016-05-17
Border Panel - Mobile Height Bug

### LINKED ISSUE #874: iOS Safari - Border Panel Height Issue
screen shot 2016-05-17 at 1 59 05 pm

kjacks:
- set border panel max height script to run when document is ready so that it can get DOM element heights

Resolves #874


sanajaved7:
Also good to merge

PR #889, opened 2016-05-18
Blog Post Header - Very Small Screens

kjacks:
- Slightly increased height of blog post header to accomodate super small screens


sanajaved7:
Also good to merge!

PR #897, opened 2016-05-23
Lighten Post Header Image Overlay

kjacks:
- Lightens image overlay for podcast, in the news, and our people templates


sanajaved7:
Good to merge 👍

PR #906, opened 2016-05-25
Added Basic Meta Tags and Social Media Tags

### LINKED ISSUE #905: Site Wide Meta Tags

kjacks:
Resolves #905

### DIFF
--- a/mysite/templates/base.html
+++ b/mysite/templates/base.html
@@ -2,6 +2,8 @@
 {% load wagtailuserbar  %}
 {% load top_menu %}
 {% load wagtailcore_tags %}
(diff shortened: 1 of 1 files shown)


mysite/templates/base.html
```diff
@@ -2,6 +2,8 @@
 {% load wagtailuserbar  %}
 {% load top_menu %}
 {% load wagtailcore_tags %}
+{% load wagtailimages_tags %}
+{% load static %}
```
sanajaved7: Hey @kjacks I don't think we need this because we have {% load static from staticfiles %} on the first line

Issue #858, opened 2016-05-16
Mobile Nav Program Logo - Mini Screens


kjacks:
adjust mobile nav section width from 50-50 to dynamic on very small screens

PR #930, opened 2016-05-27
Mini Mobile Nav Logo

kjacks:
- Set mobile nav to be 35-65% to accomodate large program logos on very small screens #858 
- Set mini program logo to not show up on pages with simple sidemenus (weekly, org-wide simple pages) #928


sanajaved7:
Also good 👍

PR #941, opened 2016-06-01
Event RSVP Time Cutoff

### LINKED ISSUE #903: RSVP link shouldn't show up after event has concluded
Easiest solution may be that the RSVP button disappears on the day of the event, but that may be too early, especially for evening events. Not sure if we can use the 'time' field to be more precise, or just have an option to remove the RSVP button (there is also the issue of some events being full, others being invite-only etc not needign RSVP buttons)

Consult with veronica + events team

kjacks:
- Added logic to factor in event time when determining whether event is future is past and to handle multi day events - Resolves #903


sanajaved7:
Looks good 👍

Issue #448, opened 2016-04-19
Editor Content Block: Button


rossvanderlinde:

rossvanderlinde:
url and text

rossvanderlinde:
left or center?

PR #949, opened 2016-06-02
Table & Button Streamfield Blocks

### LINKED ISSUE #439: Editor Content Block: Table

kjacks:
- Added table and button blocks to streamfield for post and simple page models 

Resolves #439 #448


kjacks:
This will have to wait until we upgrade to Wagtail 1.5.  Also just noting here so we remember to be careful with migrations because both the upgrade and and adding the table block involve a home migration

sanajaved7:
Looks good to merge 👍

Issue #944, opened 2016-06-02
Program Blog Home Page - Hero Grid Cutoff


kjacks:
![screen shot 2016-06-02 at 11 53 31 am](https://cloud.githubusercontent.com/assets/6720200/15751611/b4185c2c-28b8-11e6-8b7e-da8fa34fff6d.png)

PR #951, opened 2016-06-02
Program Blog Homepage Fixes

kjacks:
- fixed program blog homepage lead image overflow - #944 
- fixed program blog homepage mobile full width image error


sanajaved7:
Also good 👍

PR #967, opened 2016-06-06
Refactor Expanded Header Setup

### LINKED ISSUE #862: Header - Refactor Expanded/Desktop Header
- currently being done via js, loads desktop then expanded header

kjacks:
- Refactored home page template to explicitly set body class instead of relying on script to set class on page load - Resolves #862


sanajaved7:
Feel free to merge 👍

PR #971, opened 2016-06-06
Unify SCSS and JS Breakpoints

### LINKED ISSUE #523: Unify SCSS and JS Breakpoints
- store breakpoints in external file to be read by scss and js files, so as to limit points of change

kjacks:
- refactored js breakpoint code to draw from foundation's media query functions -> allows for both scss and js files to pull from single set of breakpoint declarations

Resolves #523


sanajaved7:
Also good to merge 👍

PR #978, opened 2016-06-07
Updating Person model queries to only return live pages not drafts

### LINKED ISSUE #855: Update Person model query
Only return live pages

sanajaved7:
[Resolves #855]


kjacks:
Also looks good!

Issue #863, opened 2016-05-17
Events - Reset Button


kjacks:
- set reset button to future events, not /events/

PR #980, opened 2016-06-08
Restructure Events Home Page Setup

### LINKED ISSUE #837: Review Events page setup
Multiple people have said they think past events are jumbled / not ordered. Actually they're ordered but start from exactly a year ago, which is too confusing.

One solution may be ordering past events (newest to oldest) differently from future events (oldest to newest). If someone selects dates that crossover both, then it can be ordered newest to oldest too.

A summer intern task could be assembling case studies of how other orgs handle events pages with past and future work. Should involve events team and Tyler B who flagged this issue

kjacks:
- implemented separate views for future and past events pages (for both org-wide and program events homepages)
- Set links in header, sidemenus, and mobile menus to new event urls

Resolves #837 #863

### DIFF
--- a/article/models.py
+++ b/article/models.py
@@ -5,7 +5,7 @@ from home.models import Post
 from wagtail.wagtailcore.models import Page
 from wagtail.wagtailadmin.edit_handlers import FieldPanel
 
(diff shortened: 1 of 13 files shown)


sanajaved7:
I think we can close this one right? Since the other branch is this one plus the tests?

PR #988, opened 2016-06-09
Tests and refactoring of new events setup

### LINKED ISSUE #837: Review Events page setup

sanajaved7:
This is looking good @kjacks! I really like your implementation here. I added the following things:
- Tests for new functionality
- Moved event specific logic under event/models.py
- Fixed merge conflicts

Resolves #837 #863

### DIFF
--- a/article/models.py
+++ b/article/models.py
@@ -5,7 +5,7 @@ from home.models import Post
 from wagtail.wagtailcore.models import Page
 from wagtail.wagtailadmin.edit_handlers import FieldPanel
 
-from mysite.helpers import paginate_results, get_posts_and_programs, get_org_wide_posts
+from mysite.helpers import paginate_results, get_program_and_subprogram_posts, get_org_wide_posts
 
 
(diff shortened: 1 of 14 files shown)


kjacks:
Looks good to merge 👍

sanajaved7:
Thanks @kjacks and before we merge, would you mind adding issues for the remaining tests to write for this feature?

sanajaved7:
@kjacks I went ahead and added the test for event ordering. Will merge now.

kjacks:
Great! Thank you so much for doing that

PR #1028, opened 2016-06-15
Post Body - Remove Body Width Cap

### LINKED ISSUE #990: Post Body - Allow media elements to break max width

kjacks:
- removes max width from post container and instead applies max width to individual block types -> allows elements like iframes, pullquotes, tables to extend beyond max body width 

Resolves #990


sanajaved7:
Good to go 👍

PR #1036, opened 2016-06-21
Home Page/Program Landing Page - Set Alternate Page Title

### LINKED ISSUE #912: Use "Page title" from promote panel for lead/feature blocks
That way we can promote content with a more descriptive title like "Watch the Highlights: Basic Event Title" to the lead/feature blocks, while keeping "Basic Event Title" in content grids etc.

Obviously would default to the usual title field (in content tab) if the page title field (in promote tab) wasn't filled in

kjacks:
Added conditionals to check for alternate title on home page and program landing pages - Resolves #912


sanajaved7:
Looks like there is a merge conflict but once that's resolved - this is good 👍

PR #1044, opened 2016-06-22
Event RSVP field change

### LINKED ISSUE #879: RSVP link in event shouldn't be compulsory
Because a few events incl. the annual conference dont have public RSVPs
Also those without an RSVP should have the RSVP button hidden. Tagging @kjacks

sanajaved7:
[Resolves #879]


sanajaved7:
Hey @kjacks FYI a program needs this today so going to merge to test on staging

kjacks:
@sanajaved7 Definitely - I do want to make sure there is enough time to test the video and iframe changes that are also in there, but also know that multiple people have been asking for the rsvp button fix

PR #1045, opened 2016-06-23
Home Page Fixes

kjacks:
- eliminated border panel info text wrapping
- set program dropdown fadeout to only appear on overflow


sanajaved7:
Also good to go 👍

PR #1054, opened 2016-06-24
Automate Events Carousel

### LINKED ISSUE #925: Automate events carousel on homepage
With next 5 events (unless there are fewer than 5)

kjacks:
- dynamically generates next five events for events carousel - Resolves #925


home/models.py
```diff
@@ -1,7 +1,11 @@
 from __future__ import unicode_literals
 
 from django.db import models
+from django.apps import apps
```
sanajaved7: Hey @kjacks I think we could import the Event model by using ```from event.models import Event``` -- or did that not work?

PR #1057, opened 2016-06-28
Header - Dynamically Populate Location Programs

kjacks:
- added context function to bring back location programs in top menu

### DIFF
--- a/home/templatetags/top_menu.py
+++ b/home/templatetags/top_menu.py
@@ -21,12 +21,12 @@ def has_menu_children(page):
 @register.inclusion_tag('tags/top_menu.html', takes_context=True)
 def top_menu(context, parent, calling_page=None):
     all_programs = Program.objects.in_menu().order_by("title").exclude(location=True)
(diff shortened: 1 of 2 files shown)


home/templatetags/top_menu.py
```diff
@@ -21,10 +21,12 @@ def has_menu_children(page):
 @register.inclusion_tag('tags/top_menu.html', takes_context=True)
 def top_menu(context, parent, calling_page=None):
     all_programs = Program.objects.in_menu().order_by("title").exclude(location=True)
+    location_programs = Program.objects.in_menu().order_by("title").filter(location=True)
 
     return {
         'calling_page': calling_page,
         'programs': all_programs,
+        'locations': location_programs,
```
sanajaved7: Hey @kjacks - minor edit but just for name consistency, we should probably call the key "location_programs" instead of "locations" just to avoid any confusion.

kjacks: @sanajaved7 sounds good - will do

PR #1065, opened 2016-06-29
Select Arrow Fix

kjacks:
- moved select arrow styling to base scss file in order to encompass all select elements


sanajaved7:
Good to merge 👍

PR #1074, opened 2016-07-06
Settings changes to fix for scheduling and file upload issues

### LINKED ISSUE #976: Add check to prevent uploading of file to Wagtail with same filename

### LINKED ISSUE #1016: Issues with Scheduling Post
Scheduled post not going live at appropriate time

sanajaved7:
These are two substantial bug fixes so adding some documentation here in case they need to be revisited at some point. 

- Solution for publishing scheduled pages - set up Heroku scheduler to run ```python manage.py publish_schedule_pages``` command every 10 minutes. **Need to relay that to comms and program staff.** This means if a page is scheduled for 9:32 a.m. - it will be published at 9:40. And likewise a page scheduled for 9:28 will go up at 9:30. cc: @rossvanderlinde @kjacks 

- Solution for file upload issue - Bug within django-storages doesn't prevent users from uploading a file with the same name. Solution in place is to prevent the s3 file from being overwritten by setting ```file_overwrite = False``` in our media custom storages file.  For users, this means if you upload a file called "OurPaper.pdf" and it changes and needs to be updated then you'll need to first delete that original file and then re-upload the new version of "OurPaper.pdf". If you were to add "OurPaper.pdf" again, it would change the file path to something like "OurPaper1234.pdf" so that it's a unique path. 

[Resolves #976] 
[Resolves #1016]

### DIFF
--- a/custom_storages.py
+++ b/custom_storages.py
@@ -7,3 +7,4 @@ class StaticStorage(S3BotoStorage):
 
 class MediaStorage(S3BotoStorage):
     bucket_name = settings.MEDIA_BUCKET_NAME
+    file_overwrite = False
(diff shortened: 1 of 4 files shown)


sanajaved7:
@kjacks let's definitely talk through this in detail today as well

kjacks:
@sanajaved7 looks great!

PR #1183, opened 2016-10-14
Set Max Width for Table Block

kjacks:
- prevents table block from breaking post layout

ex. http://www.newamerica.org/education-policy/edcentral/why-default-rates-arent-enough/


LomaxRx:
works for me

Issue #1078, opened 2016-07-11
RSS Feed


kjacks:
https://chrxr.com/adding-rss-feed-wagtail-site/

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/113920320650537/153626775468471)

rossvanderlinde:
Important RSS feed: events page.

rossvanderlinde:
![screen shot 2016-10-05 at 11 53 43 am](https://cloud.githubusercontent.com/assets/13524901/19120880/7d34f34a-8af2-11e6-9718-90955e6a2b10.png)
Updated design

LomaxRx:
current [url patterns](https://github.com/newamericafoundation/newamerica-cms/blob/17cedacfa30ac0906c5924d4c1841d602b931ed3/mysite/urls.py#L20-L25): 
**/feed/**: all content
**/feed/program/\<program-slug\>/**: all content for program
**/feed/subprogram/\<subprogram-slug\>/**: all content for subprogram
**/feed/author/\<author-slug\>/**: all content for author
**/feed/\<content-type\>/**: all content for content type (based on model name, e.g. `/feed/policypaper/`, see options [here](https://github.com/newamericafoundation/newamerica-cms/blob/17cedacfa30ac0906c5924d4c1841d602b931ed3/rss_feed/feeds.py), this is also used for events)
**/feed/\<content-type\>/\<program-slug\>/**: filter content by program (e.g `/feed/blogpost/asset-building/`)

i think these work well and cover most bases. if ya'll think so too, can move on to adding links to templates.

PR #1186, opened 2016-10-17
Rss feed

LomaxRx:
addressing #1078


rss_feed/feeds.py
```diff
@@ -0,0 +1,122 @@
...
+
+# "issueortopic", "quoted","weeklyedition", "indepthproject"
+acceptable_content_types = [
+    "book","article","blogpost",
+    "event","podcast","policypaper","pressrelease",
+    "weeklyarticle","indepthsection",
+]
+
+#"fellows","new-america-weekly",
+acceptable_programs = [
```
kjacks: @LomaxRx One thing we should consider here is the fact that the programs change pretty quickly here - so we ideally don't want to have to manually update the list each time

LomaxRx: true. i guess we can switch it to check for unacceptable programs...or not limit it at all?

kjacks: if you wanted to still limit it, you could add the line: Program.objects.live() to get all the current programs

rss_feed/feeds.py
```diff
@@ -0,0 +1,122 @@
...
+    "asset-building","better-life-lab","cybersecurity-initiative",
+    "economic-growth","education-policy","future-tense",
+    "international-security","open-technology-institute",
+    "political-reform","open-markets","resilient-communities",
+    "opportunity-at-work",
+]
+
+class GenericFeed(Feed):
```
kjacks: I'm not crazy familiar with rss feed customs, but I'm trying to think if someone would expect the generic feed to include posts and people, instead of just posts?

rss_feed/feeds.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def title(self, obj):
+        if obj["page"]:
+            return obj["page"].title
+        else:
+            raise Http404
+
+    def description(self, obj):
+        if obj["page"]:
+            return obj["page"].search_description
```
LomaxRx: this should only pull Posts, not people. 

```python
def items(self,obj):
   return Post.objects.live() ...
```
is what retrieves the posts for the feed, so no Person should be there.

rss_feed/feeds.py
LomaxRx: this should only pull Posts, not people. 

```python
def items(self,obj):
   return Post.objects.live() ...
```
is what retrieves the posts for the feed, so no Person should be there.

rss_feed/feeds.py
```diff
@@ -0,0 +1,122 @@
...
+from django.contrib.syndication.views import Feed
+from django.http import Http404
+from wagtail.wagtailcore.models import Page
+
+from home.models import Post
+from person.models import Person
+from programs.models import Program, Subprogram
+
+# "issueortopic", "quoted","weeklyedition", "indepthproject"
+acceptable_content_types = [
```
kjacks: this is a tiny thing, and feel free to keep it as is, but it might be good to keep the post type names consistent with how they appear in urls on the site (i.e. "policy-paper" instead of "policypaper") - as I said though, this could defintiely go either way

PR #1187, opened 2016-10-18
Dataviz - Download, Print, and Share Links

kjacks:
- Adds download, print, and share buttons to the dataviz block
- Buttons hidden on IE and mobile


home/templates/blocks/dataviz.html
```diff
@@ -1,23 +1,30 @@
...
+		<div class="dataviz__title-container">
+			{% if self.title %}	
+				<a id="{{ self.title|slugify }}" class="in-depth__panel__anchor"></a>
+			{% else %}
+				<a id="chart-{{ self.container_id }}" class="in-depth__panel__anchor"></a>
+			{% endif %}
+
+			<div class="dataviz__icon">{% include 'ui_elements/svg/new-america-mini.html' %}</div>
+			{% if self.show_download_link %}
+				<a class="dataviz__icon dataviz__print-link" id="{{ self.container_id }}__print-link">{% include 'ui_elements/svg/print.html' %}</a>
```
kjacks: They are handled in the visualization code [here](https://github.com/newamericafoundation/newamerica-data-projects/blob/master/src/js/viz_controller.js#L152-L162) (because it initially needed access to functions of the viz)

kjacks: though I changed the way I was initially going about it, so it might make sense to move them into the website repo

LomaxRx: it's probably fine as is, since you're doing a lot more work there, and it looks like we've been using js sparingly in the website repo. I probably would just note somewhere that that's where events are handled.

kjacks: 👍

PR #1188, opened 2016-10-19
Upgrade to wagtail 1.6.3/django 1.10

### LINKED ISSUE #1172: Upgrade Wagtail to 1.6.3
┆Issue is synchronized with this [Asana task](https://app.asana.com/0/113920320650537/188924062302309)

kjacks:
Resolves #1172 
- upgraded wagtail and its dependencies
- removed custom integer block due to new integer block added to wagtail core blocks
- changed all uses of django-overextends module to the native extends tag, now supported in django
- changed elastic search search_fields settings on post and person models from a tuple (depricated in wagtail) to a list 
- changed middleware classes (deprecated in django 1.10) to new middleware object


kjacks:
@LomaxRx After looking through release notes for wagtail 1.7, I'm thinking that we might want to hold off on upgrading just yet.  I want to give them some more time to address any bugs that the new release might have, just because some of the migrations that are required (especially for custom image model changes) seem fairly complex.  We should definitely go ahead with the update to 1.6 though, and I'd be happy to hear your thoughts!

PR #1189, opened 2016-10-19
Program Dropdown - International Reporting Project Redirect

kjacks:
- added additional exception to handle irp redirect in program dropdown


LomaxRx:
good to merge

PR #1213, opened 2016-11-15
remove overextends app

LomaxRx:
@kjacks i think this should be gone, yea? [removed overextends dependency](https://github.com/newamericafoundation/newamerica-cms/commit/606c5145b66fcc7f186ee59612a95c14d90a443a)


kjacks:
@LomaxRx yes! absolutely - I thought I had removed it in the upgrade wagtail pr, but thank you for catching that

PR #1219, opened 2016-11-23
Conferences

LomaxRx:
conferences model, template and styles.


conference/models.py
```diff
@@ -0,0 +1,119 @@
...
+from blocks import PeopleBlock, SessionsBlock, VenueBlock, DirectionsBlock, PartnersBlock
+from home.blocks import IntegerBlock
+
+class AllConferencesHomePage(Page):
+    parent_page_types = ['home.HomePage']
+    subpage_types = ['Conference']
+
+    class Meta:
+        verbose_name = "Homepage for all Conferences"
+
```
kjacks: @LomaxRx We should probably create at least a blank template at 	
conference/all_conferences_home_page.html, so that users aren't served a server error when they try to go to the conferences home page

conference/templates/conference/about.html
```diff
@@ -0,0 +1,17 @@
+{% load wagtailcore_tags wagtailimages_tags %}
+<section id="about" class="section align-center">
+  <div class="container">
+    <span class="line-icon section-icon line-icon-multimedia-12"></span>
+    <h3>About the Event</h3>
+    <div class="row">
+      <div class="small-12 medium-5 columns">
+        {% image page.about_image original as about_image %}
+        <img src="{{about_image.url}}" alt="{{about_image.alt}}" class="img-responsive" />
```
kjacks: Will there ever be conferences without an about image?  Just a heads up that I'm getting this
screen shot 2016-11-23 at 12 50 22 pm

when I don't add an image

kjacks:
@LomaxRx I love the structure you've used of keeping all the conference-specific blocks and templates in the same app directory.  For blocks like the two-column or people block that we will probably also want to use on other parts of the site, I'm wondering if it makes sense to store those block templates/models in a more central location?  I could go either way!

conference/models.py
```diff
@@ -0,0 +1,119 @@
...
+                FieldPanel('state', classname="col6"),
+                FieldPanel('zipcode', classname="col6")
+            ])
+        ],
+        heading="Conference Location"
+    )
+
+    venue = StreamField(VenueBlock(), null=True, blank=True)
+    directions = StreamField(DirectionsBlock(), null=True, blank=True)
+    speakers = StreamField(PeopleBlock(), null=True, blank=True)
```
kjacks: @LomaxRx do we want to have the ability to select people already in the new america people database here?  I know that's not super trivial to implement, so no worries either way, but it might be something for future iterations of this so that we wouldn't have to duplicate info stored about new america folks and would be able to link to a person's bio on the site easily

conference/models.py
```diff
@@ -0,0 +1,119 @@
...
+    partners = StreamField(PartnersBlock(), null=True, blank=True)
+    sessions = StreamField(SessionsBlock(), null=True, blank=True)
+
+    content_panels = [
```
kjacks: @LomaxRx this is a small thing but I'd suggest ordering the entry fields based on the order they will appear on the renderred template.  Just cuz in the past, I've found people were confused when the flow of the form didn't kind of mirror what they could expect on the site.

conference/templates/blocks/people.html
```diff
@@ -0,0 +1,23 @@
...
+      <div class="speaker" data-target="#description-{{forloop.counter}}">
+        <div class="photo-wrapper rounded"><img src="{{photo.url}}" alt="{{photo.alt}}" class="img-responsive"></div>
+        <h3 class="name">{{p.name}}</h3>
+        <p class="text-alt"><small>{{p.title}}</small></p>
+        <div id="description-{{forloop.counter}}" class="description collapse">
+          <span class="arrow"></span>
+          {{p.description}}
+        </div>
+        <ul class="speaker-socials">
+          <li><a class="icon" href="{{p.twitter}}"><span class="fa fa-twitter"></span></a></li>
```
kjacks: similarly for a speaker image as well - I know we've had problems in the past getting images for some guest authors/speakers, so I'd add a conditional here and make the image field nullable

mysite/assets/scss/modules/conference/_schedule.scss
```diff
@@ -0,0 +1,659 @@
+/* ----------------------------------------------------------------------------
+    Schedules
+---------------------------------------------------------------------------- */
+
```
kjacks: @LomaxRx suuper tiny thing, so feel free to ignore, but where you select "day 1" or "day 2" doesn't currently have a pointer cursor

conference/templates/blocks/break.html
```diff
@@ -0,0 +1,55 @@
+<section id="counters" class="section align-center overlay bg-cover bg5 light-text">
```
kjacks: @LomaxRx I'm probably missing it, but which block is this the template for?

conference/templates/blocks/location.html
```diff
@@ -0,0 +1,18 @@
+<section id="contacts">
```
kjacks: @LomaxRx I'm probably missing it, but which block is this the template for?

conference/templates/blocks/schedule.html
```diff
@@ -0,0 +1,89 @@
+<div class="container">
+    <!-- Schedule start -->
+    <div class="schedule">
+
+      {% if value|length > 1 %}
```
kjacks: @LomaxRx I probably messed this up when I was entering the info, but I wasn't able to create a single day conference

kjacks: also when I tried to add a third or fourth day, it wouldn't show up for me

conference/models.py
```diff
@@ -0,0 +1,119 @@
+import json
+import datetime
+
```
kjacks: @LomaxRx i might suggest removing any of these imports if you don't end up using them, just for performance/cleanliness stuffs - though probably negligible anyways

conference/templates/blocks/schedule.html
```diff
@@ -0,0 +1,89 @@
...
+        {% for day in value %}
+          <li data-target="#day{{forloop.counter}}">
+            <span>
+              <h5 class="highlight">Day {{forloop.counter}}</h5>
+              {% comment %}
+                we do not have access to main template's context in this template
+                to use / format conference dates
+                ...when we upgrade to wagtail 1.7 we can use 'include_block' tag which will
+                give us access to parent template's context
+              {% endcomment %}
```
kjacks: awesome :)

conference/templates/conference/conference.html
```diff
@@ -0,0 +1,68 @@
+{% extends 'conference/base.html' %}
```
kjacks: see comment above

conference/templates/conference/intro.html
```diff
@@ -0,0 +1,40 @@
+{% load wagtailcore_tags %}
+{% load wagtailimages_tags %}
+
+{% image page.story_image original as cover %}
+<section id="hero" style="background-image:url({{cover.url}});" class="hero-section bg1 bg-cover window-height light-text">
+
+  <div class="heading-block centered-block align-center">
+    <div class="container">
+      <div class="na-logo">
+        <a href="//newamerica.org/" target="_blank">{% include 'ui_elements/svg/new-america-white.html' %}</a>
```
kjacks: @LomaxRx -> "https://"

mysite/assets/scss/base/type/_lineicons.scss
```diff
@@ -0,0 +1,1874 @@
...
+}
+.line-icon-shopping-12:before {
+  content: "\e180";
+}
+.line-icon-shopping-13:before {
+  content: "\e181";
+}
+.line-icon-shopping-14:before {
+  content: "\e182";
+}
```
kjacks: @LomaxRx Ideally if possible we could add this to the font static file on s3 like the other fonts

conference/templates/conference/about.html
LomaxRx: veronica said she was fine always using an image there, but it's trivial to make an exception, so i'll add this

conference/models.py
LomaxRx: i think this will be a good feature to think about for the future, but for now, in the interest of time, i'm just going to keep it as is.

conference/models.py
LomaxRx: my thought was to put the longer sections at the bottom of the input (so staff wouldn't have to scroll through 20 speakers/30 sessions to update the directions. I also couldn't see a way to make a single StreamField collapsible in the input...did I miss something? but i think either way is ok, unless there's a way to make it collapsible, then doing it in order is a much better option

conference/templates/blocks/schedule.html
LomaxRx: For now, we're only going to support 2 days...since veronica said we'll ***never*** have more than a 2 day conference. Would you be concerned about user error? Or maybe I should limit the options to just Day 1 and Day 2

conference/templates/conference/base.html
```diff
@@ -0,0 +1,73 @@
...
+    </head>
+
+    <body class="{% block body_class %}{% endblock %}">
+        {% wagtailuserbar %}
+
+          {% block content %}
+          {% endblock %}
+
+    </body>
+</html>
```
LomaxRx: i didn't want to import the NA sitewide header. I guess I could just use that template and simply hide it in the conference template.

conference/models.py
kjacks: Sounds good!

conference/templates/blocks/schedule.html
kjacks: Since it's only a select number of people who are going to be creating and editing these, I wouldn't worry about it too much, but if it's an easy fix then maybe making it just 2 days for now might make sense

mysite/assets/js/utilities/scrollr.js
```diff
@@ -0,0 +1,122 @@
...
+    return this.window.currentPosition >= this.scrollTop();
+  }
+
+  hasLeft(){
+    return this.window.currentPosition >= this.scrollTop() + this.element.outerHeight();
+  }
+}
+
+const scrollr = new Scrollr();
+export default scrollr;
```
kjacks: that's awesome :)

conference/templates/conference/base.html
kjacks: yeh, if it's possible to make that change I feel like it would be ideal, just so that if we have to make edits to the base template in the future we won't have to remember to make them in two places

conference/templates/conference/intro.html
LomaxRx: [https](https://newamerica.org) does not seem to be implemented

Issue #1206, opened 2016-11-03
Tracking downloads on Google Analytics


LomaxRx:
how many people have downloaded their paper from our website.
[Reference](https://support.google.com/analytics/answer/1012044?hl=en)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/113920320650537/207046215007169)

LomaxRx:
@rossvanderlinde @kjacks i'm thinking we should switch to [google tag manager](https://www.google.com/analytics/tag-manager/features/). It makes it much easier to add additional analytics to site without have to alter script

PR #1220, opened 2016-11-23
swapping google analytics for google tag manager

LomaxRx:
addresses #1206

we will also need to set up google [tag manager events](https://tagmanager.google.com/?hl=en#/container/accounts/929557346/containers/6052491/workspaces/1).

I'm going to see if I can set that up before we push this.


LomaxRx:
@kjacks, i set up google tag manager and tested.

PR #1230, opened 2016-12-07
Conferences

LomaxRx:
fixes per conversation with Ellen


conference/blocks.py
```diff
@@ -13,8 +13,8 @@
 
 class PersonBlock(blocks.StructBlock):
     name = blocks.TextBlock(required=True)
-    title = blocks.TextBlock()
-    description = blocks.RichTextBlock()
+    title = blocks.TextBlock(required=False)
+    description = blocks.RichTextBlock(reqired=False)
```
kjacks: reqired => required

LomaxRx: ouch! nice catch

Issue #502, opened 2016-04-21
Story Excerpt Help Text Needed


rossvanderlinde:
We need to tell people what the character count is inside the editor.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/113920320650537/115014319583076)

rossvanderlinde:
And probably the purpose of the field too -- the fact that it populates summaries

PR #1233, opened 2016-12-13
Help Text for Story Excerpt

LomaxRx:
Addresses #502


home/models.py
```diff
@@ -418,7 +418,7 @@ class Post(Page):
     post_author = models.ManyToManyField(
         Person, through=PostAuthorRelationship, blank=True)
 
-    story_excerpt = models.CharField(blank=True, null=True, max_length=140)
+    story_excerpt = models.CharField(blank=True, null=True, max_length=140, help_text="140 character limit. If not 'Search Description' is defined, this populates the post summary that shows up on search results on newamerica.org and is indexed by google")
```
kjacks: @LomaxRx maybe changed to "140 character limit. Populates the post summary that shows up in search results on newamerica.org site pages.  Serves as default search description to be indexed by google if 'search_description' field is undefined" - because either way, the story excerpt is used on the internal site no matter what is in the search description field

LomaxRx:
not going to bother with this

Issue #1218, opened 2016-11-18
301 redirect from http to https


rschulman:
Can we please implement a permanent 301 redirect from the insecure version of our site to the https version? Since we already serve https there is no reason I can think of to continue serving the insecure pages. A simple 301 from any page at http to the same page at https is easy and will do the job.

In the future, once the 301 is implemented and stable, I would recommend also turning on HTTP Strict Transport Security to prevent man in the middle TLS stripping attacks.

@georgiamoon said that tagging @LomaxRx would be helpful, so there we go.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/113920320650537/224288552329661)

rossvanderlinde:
Thanks @rschulman we'll get working on this.

PR #1234, opened 2016-12-14
SSL redirect in Django Settings

LomaxRx:
addresses #1218 

@rschulman , looks like a simple [Django settings tweak](https://docs.djangoproject.com/en/dev/ref/settings/#secure-ssl-redirect) is all we need. 

The [SECURE_PROXY_SSL_HEADER](https://docs.djangoproject.com/en/dev/ref/settings/#secure-proxy-ssl-header) is apparently necessary for Heroku instances...but Django docs warn "Obviously, you should only set this setting if you control your proxy or have some other guarantee that it sets/strips this header appropriately." 

Here's Heroku's [documentation](https://devcenter.heroku.com/articles/http-routing#routing) on how they route requests, so it looks like we can reasonably trust that the header is being set appropriately by Heroku... are there any other considerations?

### DIFF
--- a/mysite/settings/production.py
+++ b/mysite/settings/production.py
@@ -18,6 +18,9 @@ ADMINS = [('Kirk', 'jackson@newamerica.org'), ('Andrew', 'lomax@newamerica.org')
 # Will be changed to final host url
 ALLOWED_HOSTS = ['*']
 
+SECURE_SSL_REDIRECT = True
+SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
+
 AWS_QUERYSTRING_AUTH = False
 
 # s3 bucket settings


rschulman:
Just to make sure I'm not giving the wrong advice, is it correct that the following is how our service is currently provisioned?:
- Django is our content management service, and it responds to http(s) requests.
- Heroku is our web server, it does not do much of anything other than forward http(s) requests to Django. Particularly, it doesn't terminate/load balance incoming requests?

If that's right, I believe that yes, the change you're contemplating here looks like it should work.

LomaxRx:
Correct, we use Django and Heroku. However, Heroku _does_ use load balancers and terminates SSL requests, but it looks like it also, based on the header of the incoming request to the load balancer, resets the X-Forwarded-Proto header when routing 

> ## [Heroku headers](https://devcenter.heroku.com/articles/http-routing#heroku-headers)
> All headers are considered to be case-insensitive, as per HTTP Specification.
> - X-Forwarded-For: the originating IP address of the client connecting to the Heroku router
> - **_X-Forwarded-Proto: the originating protocol of the HTTP request (example: https)_**
> - X-Forwarded-Port: the originating port of the HTTP request (example: 443)
> - X-Request-Start: unix timestamp (milliseconds) when the request was received by the router
> - X-Request-Id: the Heroku HTTP Request ID
> - Via: a code name for the Heroku router

Issue #762, opened 2016-05-10
Resizing images after they've been placed into CMS


rossvanderlinde:
from Maria: I added an image to a simple page, but when it placed it in, it came in at full size. Being able to resize and/or move images around in the CMS would be a helpful feature in the future.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/113920320650537/131722831128264)
┆Attachments: <a href="https:&#x2F;&#x2F;app.asana.com&#x2F;app&#x2F;asana&#x2F;-&#x2F;download_asset?asset_id&#x3D;131722831128270">image size.PNG</a>

LomaxRx:
We can add an image width field to allow users to specify % for rich text block. but not for image block.

PR #1236, opened 2016-12-15
Custom Image Block for Inline Images

LomaxRx:
Depends on: #1235
Addresses: #762

Should most always be used in place of the old image block. 

- Gives editor option for left, right or full-width alignment
- Options for Auto, 60%, 50%, 33% or 25% of post's width
- Modifies image block template to take either a wagtail `image` object (with source and caption fields) or the value of the `CustomImageBlock` object
- Restyles inline images inside of an image wrapper
- Centers inline images on mobile


LomaxRx:
I'm thinking we should relabel the block in the editor from 'Custom Image' to 'Inline Image'

kjacks:
agreed 👍  - I feel like "inline" will make it even more clear/specific

Issue #1244, opened 2017-01-24
KeyError in search results/ elasticsearch


LomaxRx:
```sh
Template error:
In template /app/search/templates/search/search.html, error at line 59
   fields
   49 :
   50 :                     {% if search_results.has_next %}
   51 :                         <a class="pagination__arrow" href="{% url 'search' %}?query={{ search_query|urlencode }}&page={{ search_results.next_page_number }}">
   52 :                             {% include 'ui_elements/svg/arrow-right.html' %}
   53 :                         </a>
   54 :                     {% endif %}
   55 :                 </div>
   56 :             </div>
   57 :         </div>
   58 :
   59 :          {% if search_results %}
   60 :             <div class="content-grid__entries">
   61 :
   62 :             {% for post in search_results %}
   63 :
   64 :                 {% if post.specific.content_type|stringformat:"s" == "person"  %}
   65 :                     <div class="content-entry">
   66 :                         {% if post.specific.profile_image %}
   67 :                             <div class="post-person with-image">
   68 :                                 <a href="{{ post.specific.url }}">
   69 :                                     <div class='person__icon'>


Traceback:

File "/app/.heroku/python/lib/python2.7/site-packages/django/core/handlers/exception.py" in inner
  39.             response = get_response(request)

File "/app/.heroku/python/lib/python2.7/site-packages/django/core/handlers/base.py" in _get_response
  187.                 response = self.process_exception_by_middleware(e, request)

File "/app/.heroku/python/lib/python2.7/site-packages/django/core/handlers/base.py" in _get_response
  185.                 response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "/app/search/views.py" in search
  39.         'search_results': search_results,

File "/app/.heroku/python/lib/python2.7/site-packages/django/shortcuts.py" in render
  30.     content = loader.render_to_string(template_name, context, request, using=using)

File "/app/.heroku/python/lib/python2.7/site-packages/django/template/loader.py" in render_to_string
  68.     return template.render(context, request)

File "/app/.heroku/python/lib/python2.7/site-packages/django/template/backends/django.py" in render
  66.             return self.template.render(context)

File "/app/.heroku/python/lib/python2.7/site-packages/django/template/base.py" in render
  208.                     return self._render(context)

File "/app/.heroku/python/lib/python2.7/site-packages/django/template/base.py" in _render
  199.         return self.nodelist.render(context)

[...]

PR #1253, opened 2017-01-25
Downgrade Elasticsearch-py

LomaxRx:
addressing #1244

Can't think of a way to test this outside of production since error is not replicable in staging or locally. Downgrading elasticsearch to match Searchbox's version.


LomaxRx:
not going to do this.

PR #1265, opened 2017-02-08
Webpack upgrade

kjacks:
- upgrade to Webpack 2

@LomaxRx - if you have time at some point to try running this locally and verify that it is working.  They changed a ton from version 1 to 2, and I think I hit all the bases, but I'd love to verify


LomaxRx:
for sure

Issue #1201, opened 2016-10-25
Production debugging / logging solution


LomaxRx:
some reference: [Django](http://docs.djangoproject.com/en/dev/topics/logging)
[Wagtail](http://docs.wagtail.io/en/v1.6.3/advanced_topics/settings.html)
[Heroku](https://elements.heroku.com/addons#logging)

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/113920320650537/202458528104289)

PR #1280, opened 2017-03-01
Logging

LomaxRx:
Error logging with [LogDNA API](https://github.com/logdna/python) addressing #1201

writes error like:
`Mar 1 10:31:09 <hostname> newamerica-cms [ERROR] file=<filepath> line=<lineno> msg="<message>"`

in LogDNA console, full traceback available in context object. 
can now add environment variable HOSTNAME to Heroku configs


LomaxRx:
@kjacks, I changed this from a logging setting to middleware so we could capture more context from the request.

PR #1307, opened 2017-05-02
In Depth Profile Page

kjacks:
- Enables users to create a page within an in depth project that aggregates data for an individual data point (i.e. a state profile page that has fields from the live data source for that state)


LomaxRx:
@kjacks I pulled this down because I couldn't quite figure out what it was doing. When I went to create an in-depth Profile page in the editor under an existing project, I got a 404 error  
```
Request URL: http://localhost:8000/admin/pages/add/in_depth/indepthprofile/12456/
Raised by:	wagtail.wagtailadmin.views.pages.create
```
It's possible that my database is jacked up because I didn't roll back migrations I've made on other branches, but wanted to make sure this is an issue with my local environment only.

kjacks:
@LomaxRx I'm not getting the error on my end - I'll try dropping my local db and running the migration again just to be sure

LomaxRx:
ok, i'll do the same

LomaxRx:
yup, fixed it! false alarm.

kjacks:
great! Let me know if you want me to walk through it - cuz I know it's not all that intuitive

PR #1345, opened 2018-10-17
Cross browser report

LomaxRx:
- fixes bottom chapter-nav alignment in firefox
- fixes endnote transition in safari


lorenries:
lit

PR #1355, opened 2018-11-15
update eleastic search version

genuineBuildMonkey:
update to elastic search 6+

### DIFF
--- a/newamericadotorg/settings/production.py
+++ b/newamericadotorg/settings/production.py
@@ -59,7 +59,7 @@ es_url = os.getenv('SEARCHBOX_URL', "http://localhost:9200/")
 
 WAGTAILSEARCH_BACKENDS = {
     'default': {
-        'BACKEND': 'wagtail.search.backends.elasticsearch2',
+        'BACKEND': 'wagtail.search.backends.elasticsearch6',
         'URLS': [es_url],
         'INDEX': 'elasticsearch',
         'TIMEOUT': 1500,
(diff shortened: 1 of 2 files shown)


lorenries:
Thanks Paul,

[Looking at the Wagtail docs](http://docs.wagtail.io/en/v2.3/topics/search/backends.html#elasticsearch-backend), if we bump to elastic search 6+, we need to use 6.3.0 or earlier, and also update the Wagtail settings [here](https://github.com/newamericafoundation/newamerica-cms/blob/c1e10a42fc7d6a42eb08d75acd38db55de418376/newamericadotorg/settings/production.py#L62).

genuineBuildMonkey:
It looks like the extra keys in settings are optional?


> WAGTAILSEARCH_BACKENDS = {
>     'default': {
>         'BACKEND': 'wagtail.search.backends.elasticsearch2',
>         'URLS': ['http://localhost:9200'],
>         'INDEX': 'wagtail',
>         'TIMEOUT': 5,
>         'OPTIONS': {},
>         'INDEX_SETTINGS': {},
>     }
> }
> 
> Other than BACKEND, the keys are optional and default to the values shown.

..Which is not to say we won't have to tweak a setting somewhere.

Can we try 6.3.0 and see what it does?

genuineBuildMonkey:
Thanks. Didn't get the relevance of the 2.

lorenries:
yeah but I think we also have to change `BACKEND` to `wagtail.search.backends.elasticsearch6`, everything else is optional. Ok, merging, 🤞.

PR #1356, opened 2018-11-15
comment out __str__ overide method in home/models in attempt to get c…

genuineBuildMonkey:
https://stackoverflow.com/questions/39300060/authorizationexception-exception-str-failed-on-scan

we aren't able to see actual exception from indexing attempts. commenting out string override method in models to view exception.


genuineBuildMonkey:
Problem resolved. closing this.

PR #1357, opened 2018-11-17
require name input rather than providing on file upload

genuineBuildMonkey:
Overwrite multiple image add template to blank text input on file upload forcing users to enter custom name.

### DIFF
--- a/newamericadotorg/templates/wagtailimages/multiple/add.html
+++ b/newamericadotorg/templates/wagtailimages/multiple/add.html
@@ -1,0 +1,113 @@
+{% extends "wagtailimages/multiple/add.html" %}
+
+{% load i18n %}
+{% load l10n %}
+{% load staticfiles %}
+{% load wagtailimages_tags %}
+{% block titletag %}{% trans "Add multiple images" %}{% endblock %}
+
+{% block content %}
+    {% trans "Add images" as add_str %}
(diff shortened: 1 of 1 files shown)


genuineBuildMonkey:
This was done in HTML and JS rather than attempting to modify "form" or internal wagtail data queries because those possibly (likely) are re-used when this form is presented filled later.

This rewrites the wagtail template for image upload similar to these (one of which also includes embedded script): https://github.com/newamericafoundation/newamerica-cms/tree/master/newamericadotorg/templates/wagtailadmin.

The mode of action for that particular Wagtail page is to generate form HTML for upload form and return it as an HTML snippet to an Ajax query. Then it uses jQuery to insert the form HTML into the page. This modification overwrites the "done" callback from Wagtail to remove data from the form before inserting into the page.

https://github.com/newamericafoundation/newamerica-cms/blob/5a6f885c87b923425f8063085f905a54cc09d8fd/newamericadotorg/templates/wagtailimages/multiple/add.html#L92

Hope this helps explain the how/why. jQuery and Ajax are already in use by the Wagtail template. Template modifications are similar to ones that already exist. Deeper modifications might have side effects in other areas.

genuineBuildMonkey:
Closed over issues with maintainability in altering core wagtail functionality.

PR #1369, opened 2019-04-25
Issue/1367 provide iframe fallback image

kevinhowbrook:
Fixes #1368 

Add extra fields to the iframe block so when iframes are rendered for the PDF, images are used as fallbacks.

This can be tested on stage here in the first streamfield block under the #sections tab https://na-staging.herokuapp.com/education-policy/reports/undermining-pell-iv/

The PDF result can be seen here https://na-staging.herokuapp.com/education-policy/reports/undermining-pell-iv/pdf/

example iframe on pdf


report/templates/report/figure_iframe.html
```diff
@@ -0,0 +1,20 @@
+{% load utilities %}
+<figure class="figure {{value.fallback_image_align}} {{value.fallback_image_width|temp_image_width_map}}">
+  <div class="figure__image">
+      <img src="https://s3.amazonaws.com/newamericadotorg/{{value.fallback_image.file.name}}"/>
```
kaedroho: Any way we can get this URL from the storage backend, or make the s3 bucket bas URL configurable using a setting instead?

newamericadotorg/blocks.py
```diff
@@ -62,10 +62,24 @@ class Meta:
...
 class IframeBlock(blocks.StructBlock):
 	source_url = blocks.URLBlock(required=True)
 	width = IntegerBlock(max_value=1050, help_text="The maximum possible iframe width is 1050")
 	height = IntegerBlock()
+	fallback_image = ImageChooserBlock(icon="image", required=False,  help_text="The fallback image will be rendered for the PDF")
+	fallback_image_align = blocks.ChoiceBlock(choices=[
+		('center', 'Centered'),
+		('left', 'Left'),
+		('right', 'Right')
+	], default='center', required=False)
```
kaedroho: As there's a  default, these fields probably should be required.

report/templates/report/figure_iframe.html
kevinhowbrook: Thanks, I've done this with a template tag so it can be re-used https://github.com/newamericafoundation/newamerica-cms/pull/1369/commits/d0bd457010d5b521e09376c7cd4c3d052fcd5e1a

newamericadotorg/blocks.py
kevinhowbrook: Thanks, added a migration to correct this

report/templates/report/figure_iframe.html
kevinhowbrook: I've had to revert this, after re-testing on stage the image doesn't render. Also MEDIA_URL for settings isn't used anywhere in the codebase so I suspect there is another reason absolute urls are being used

PR #1376, opened 2019-06-10
1370 iframe sizing

nmorduch:
- Add a column width to iframe blocks. Only changes the template if that width is set, so shouldn't cause problems with existing code.
- Add a streamfield block for Datawrapper embeds. Currently takes the full embed code, but also takes the source url because we may want to start using the Datawrapper API instead in the future.

TODO (for me):
- Is the js in the datawrapper embed code not resizing the iframe because of my browser settings or because of something else?
- Check that this hasn't messed up older pages
- Check that the new block is available places like blog posts as well as reports
- Is it the source url we want to be collecting or just the 5 letter code for a given datawrapper chart? Likely the latter tbh

@kevinhowbrook Even though I have some things to do, I think it would still be helpful to review this now so that when I make any upcoming changes I can also fix anything you find. Thank you :)


newamericadotorg/templates/blocks/iframe.html
```diff
@@ -2,8 +2,15 @@
 {% load noShowTableauHome from utilities %}
 
 {% widthratio self.height self.width 100 as height_width_ratio %}
```
kevinhowbrook: I've ever seen this before, what is the `100` doing here?

kevinhowbrook:
Thanks @nmorduch, LGTM so far. 
Could you regenerate [the migration](https://github.com/newamericafoundation/newamerica-cms/blob/9f83f722a2c9f62b9214f11672d3488906eb7a19/report/migrations/0020_auto_20190607_1047.py) with a nake though instead of `auto` ?

eg `python mange.py makemigrations -n add_data_wrapper_block`

This makes it easier to understand what a migration file is doing without digging into the code.

Cheers

nmorduch:
Putting this on staging, will do my further testing there.

newamericadotorg/templates/blocks/iframe.html
nmorduch: Forgot to answer! I have no idea, didn't touch it. But my guess is that widthratio divides height and width, and 100 scales it to a percentage, or something like that.

PR #1382, opened 2019-06-20
Remake Datawrapper block migration migratation

nmorduch:
Changed because of updated help text


nmorduch:
Wow typo. Noticed that a little late...

PR #1386, opened 2019-06-28
Initial work on datawrapper tabs on funding page

nmorduch:
@kevinhowbrook Can you take a look at this? I'm having trouble figuring out how to make it so that going directly to http://localhost:8000/our-funding/2018 doesn't 404. React-router doesn't seem to like to do hashes unless you're using the hash router, and I didn't find where we're telling Django that the urls like our-funding/our-funders should be handled by react.

To see my changes, go to  http://localhost:8000/our-funding/ and see the datawrapper embed (currently just example things) as well as the nav to switch between the 2019 and 2018 table (or, right now, the table about moore's law and the chart about voting).

Note that the datawrapper embeds don't have responsive height because the JS for that is on the other branch (see #1385). 

image

Changes:
- Use Datawrapper chart instead of custom chart
- Use existing tab setup to allow multiple charts
- Currently the url /our-funding/2018 will 404 if you refresh it, because Django then handles it

### DIFF
--- a/newamericadotorg/assets/react/home-panels/pages/OurFunding.js
+++ b/newamericadotorg/assets/react/home-panels/pages/OurFunding.js
@@ -232,11 +232,11 @@ class OurFundingMain extends Component {
           </div>
           <div className="dataviz__chart-container">
             <Switch>
-              <Route exact path="/our-funding/" render={(props)=>(
-                <iframe title="Moore's Law: &quot;The transistor count doubles every two years.&quot;" aria-label="Table" id="datawrapper-chart-UOHEE" src="//datawrapper.dwcdn.net/UOHEE/1/" scrolling="no" frameborder="0" width="100%" height="600"></iframe>
-              )}/>
               <Route exact path="/our-funding/2018/" render={(props)=>(
                 <iframe title="European countries with the lowest &amp;amp; highest voter turnout" aria-label="Bar Chart" id="datawrapper-chart-vNnBJ" src="//datawrapper.dwcdn.net/vNnBJ/1/"  scrolling="no" frameborder="0" width="100%" height="650"></iframe>
+              )}/>
+              <Route render={(props)=>(
+                <iframe title="Moore's Law: &quot;The transistor count doubles every two years.&quot;" aria-label="Table" id="datawrapper-chart-UOHEE" src="//datawrapper.dwcdn.net/UOHEE/1/" scrolling="no" frameborder="0" width="100%" height="600"></iframe>
               )}/>
             </Switch>
       		</div>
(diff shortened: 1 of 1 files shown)


newamericadotorg/assets/react/home-panels/pages/OurFunding.js
```diff
@@ -4,7 +4,7 @@ import React, { Component } from 'react';
 import ImageAside from '../components/ImageAside';
 import Reel from '../components/Reel';
 import Body from '../components/Body';
-import { Route, NavLink } from 'react-router-dom';
+import { Route, Switch } from 'react-router-dom';
```
nmorduch: NavLink wasn't actually being used on the page. I started using Switch when I was reading about react-router. I don't think we _need_ to use it but it seems appropriate to, as far as I can tell.

newamericadotorg/assets/react/home-panels/pages/OurFunding.js
```diff
@@ -247,22 +260,24 @@ export default class OurFunding extends Component {
...
-        <Route path="/our-funding/corporate-circle/" render={(props)=>( <FunderLists data={this.findSubpage('corporate-circle').data} /> )}/>
-        <Route path="/our-funding/new-america-councils/" render={(props)=>( <Councils data={this.findSubpage('new-america-councils').data} /> )}/>
-        <Route path="/our-funding/donate/" exact render={(props)=>(
-          <div className="container--1080 margin-80">
-            <div className="row gutter-20">
-              <article className="col-md-7 post-body home__panel__body__text">
-              <div className="margin-bottom-35">
-                <a className="button" href={donate.data.donate.button[0].button_link}>{donate.data.donate.button[0].button_text}</a>
+        <Switch>
+          <Route exact path={["/our-funding/", "/our-funding/2018/"]} render={(props)=>( <OurFundingMain data={results.data} dataScript={results.data_project_external_script}/> )}/>
```
nmorduch: I'm not seeing highlighting about this but I made them all use exact. I'm thinking now that since not matching any of them won't make it 404, it would be better to use the pattern where this line is at the end without a path set, so that it's the default. Though also I think it would be good to figure out a 404 setup.

newamericadotorg/assets/react/home-panels/pages/OurFunding.js
```diff
@@ -223,9 +223,22 @@ class OurFundingMain extends Component {
...
+            <HorizontalNav className={`our-funding__nav`} items={[
+              { url: '/our-funding/', label: '2019', exact: true },
+              { url: '/our-funding/2018/', label: '2018'},
+            ]} />
+          </div>
           <div className="dataviz__chart-container">
-      			<div className="dataviz__chart-area" id={transparency_table.dataviz[0].container_id}></div>
+            <Switch>
+              <Route exact path="/our-funding/" render={(props)=>(
+                <iframe title="Moore's Law: &quot;The transistor count doubles every two years.&quot;" aria-label="Table" id="datawrapper-chart-UOHEE" src="//datawrapper.dwcdn.net/UOHEE/1/" scrolling="no" frameborder="0" width="100%" height="600"></iframe>
```
nmorduch: These particular charts are just for testing. I don't want to put in the real ones until they're finalized.

nmorduch:
Making a redirect in the admin does indeed work. Gonna put this on staging, and am ready for a review!

nmorduch:
To give a more detailed explanation:

For now, we're just going to have a redirect from '/our-funding/2018' to '/our-funding'. It's not an amazing solution, but is quick, and makes sense since these pages are going to be refactored anyway.

nmorduch:
oh no, this was supposed to be into staging. gonna revert.

PR #1388, opened 2019-07-01
Feature: funding table

nmorduch:


nmorduch:
Now merging into staging, which I meant to do with #1386

PR #1392, opened 2019-07-15
WIP: Funding table changes (branch is `staging-test`)

nmorduch:
So having accidentally merged to master and then reverted messed with the diffs. This branch is `staging` except I rebased and squashed the first two commits about the funding table together. I apparently can't push to staging, though.

This is waiting for Development to be ready, before we merge it.

### DIFF
--- a/newamericadotorg/assets/react/home-panels/pages/OurFunding.js
+++ b/newamericadotorg/assets/react/home-panels/pages/OurFunding.js
@@ -4,7 +4,7 @@ import React, { Component } from 'react';
 import ImageAside from '../components/ImageAside';
 import Reel from '../components/Reel';
 import Body from '../components/Body';
-import { Route, NavLink } from 'react-router-dom';
+import { Route, Switch } from 'react-router-dom';
 import HorizontalNav from '../../components/HorizontalNav';
 
 class FunderList extends Component {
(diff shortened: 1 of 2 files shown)


newamericadotorg/assets/react/home-panels/pages/OurFunding.js
```diff
@@ -223,9 +223,20 @@ class OurFundingMain extends Component {
         </div>}/>
         <div className="funding-table container--1080">
           <h1>{transparency_table.heading[0]}</h1>
-          <h3>{transparency_table.dataviz[0].title}</h3>
+          {/* <h3>{transparency_table.dataviz[0].title}</h3> */}
```
kevinhowbrook: Is this meant to be commented out? Just checking as it seems like an import `H` tag.

newamericadotorg/assets/react/home-panels/pages/OurFunding.js
```diff
@@ -247,22 +258,24 @@ export default class OurFunding extends Component {
...
+          <Route exact path="/our-funding/corporate-circle/" render={(props)=>( <FunderLists data={this.findSubpage('corporate-circle').data} /> )}/>
+          <Route exact path="/our-funding/new-america-councils/" render={(props)=>( <Councils data={this.findSubpage('new-america-councils').data} /> )}/>
+          <Route exact path="/our-funding/donate/" render={(props)=>(
+            <div className="container--1080 margin-80">
+              <div className="row gutter-20">
+                <article className="col-md-7 post-body home__panel__body__text">
+                <div className="margin-bottom-35">
+                  <a className="button" href={donate.data.donate.button[0].button_link}>{donate.data.donate.button[0].button_text}</a>
+                </div>
+                <div dangerouslySetInnerHTML={{__html: donate.data.donate.paragraph[0] }} />
```
kevinhowbrook: `dangerouslySetInnerHTML` is a bit new to me, but i'm of the opinion we can trust the value we are using because it's coming from the cms data, if not then I think we would need to send this value through a sanitizer.

newamericadotorg/assets/react/home-panels/pages/OurFunding.js
nmorduch: Yeah, we're not going to use it but I thought I'd leave it just in case we wanted it back later, or something? I think it mostly just doesn't matter since this is going to get redone

newamericadotorg/assets/react/home-panels/pages/OurFunding.js
nmorduch: I didn't write this (just moved these lines around) and I def want to look into it a little more when it comes time to redo the page.

PR #1408, opened 2019-09-20
PDF generator with external service

kaedroho:
This moves the PDF generation into an external service allowing the memory to be reduced on production.

The service is a flask app that takes the report HTML, runs it through Weasyprint and uploads the generated PDF directly to S3 using a pre-signed upload URL. I currently have an instance of that running on Google Cloud Run which could be used for testing (please DM me for URL).


nmorduch:
I am not an appropriate reviewer for this one. @AnnaGrieve, can we get a BE dev to review this? Thank you!

nmorduch:
I'm also noting here that we'll want to update the wiki with information about this change. Probably on this page: https://github.com/newamericafoundation/newamerica-cms/wiki/PDF-Generation

AnnaGrieve:
Matt to review and Jacob to update wiki - need to check these tow have access.

report/tasks.py
```diff
@@ -1,37 +1,43 @@
...
-    revision.save_revision()
+    doc.file.save(revision.title + last_edited, BytesIO())
+
+    # Submit request to PDF generator
+    upload_successful = render_pdf(html, doc.file.name)
+
+    if upload_successful:
+        revision.report_pdf = doc
+        revision.generate_pdf_on_publish = False
+        revision.save_revision()
```
gasman: Probably makes sense to have an `else:` clause to delete the `Document` record (and corresponding empty file) if the render task fails...?

AnnaGrieve:
Hi @kaedroho looking at @nmorduch's comment, OK to deploy this to stage as discussed this morning.

nmorduch:
@kaedroho is this ready to test on staging? and how should we go about doing so?

AnnaGrieve:
Hi @nmorduch, yes this is ready to test on stage

AnnaGrieve:
Hi @nmorduch - can you confirm if this change has been deployed to your production environment, or is this something you are expecting us to do?

nmorduch:
It has been deployed.

PR #1409, opened 2019-10-04
Move generated base.html file into separate templates folder

kaedroho:
Also rename the templates/index.html to templates/base.html.


nmorduch:
Also noting here for posterity that this was checked on staging as well.

PR #1404, opened 2019-09-11
1399 tweak ogimage size

victoriachan:
Scale down og and twitter image sizes. See issue [1399](https://github.com/newamericafoundation/newamerica-cms/issues/1399)

PR #1411, opened 2019-10-04
Scale down og and twitter images

nmorduch:


nmorduch:
Fun trivial merge conflicts, thanks git.

nmorduch:
This no longer has the whitespace changes but is otherwise identical to #1404 so I'm calling it not-in-need-of-further-review

Issue #1399, opened 2019-08-19
[<1hr] Scale down open graph images


nmorduch:
Twitter often doesn't show image previews for reports. I believe the cause of this is that we're using the original images for open graph images. Twitter supports "minimum dimensions of 300x157 or maximum of 4096x4096 pixels. Images must be less than 5MB in size." ([source](https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/summary))

For what to scale them to, I think this (found via a quick search) is a pretty good survey of the existing recommendations: https://iamturns.com/open-graph-image-size/

I'm inclined to do: if `(width / height) > 1.45` use `fill-1200x630`, otherwise use `fill-1200x1200`. Since 1200x630 is a ratio of about ~1.9:1, this puts the cutoff halfway between that and square.

mariaelkin:
I would say this is a high priority fix for torchbox.

kevinhowbrook:
Thanks for the information @nmorduch 

Looks like this is a case of correcting the [`base.html`](https://github.com/newamericafoundation/newamerica-cms/blob/master/newamericadotorg/templates/base.html#L36) to use an image tag with the correct size

E.g:

```
{% if page.story_image  %}
    {% image page.story_image width-320 as story_image %}
     <meta name="twitter:image" content="{{ story_image.url }}">
{% else %}
...
```

I would estimate this up to an hours work

victoriachan:
Hi Naomi,

It isn't straightforward to do math operations on the template. We normally use `width-320` for the open graph images. Do you want me to use this or an alternative resize rule that doesn't require math operations?

Regards,
Victoria

nmorduch:
Oh right. Let's go with `fill-1200x630` and I'll add some math someday when I have free time.

victoriachan:
Hi Naomi, 

I've created an PR for this now: https://github.com/newamericafoundation/newamerica-cms/pull/1404 If you're happy with that I can put it on staging for testing. Though you probably wouldn't be able to test on Twitter Card validator etc if the site requires http authentication.

Regards,
Victoria

nmorduch:
Merged but putting in `In review` because I want to deploy to and test in production before closing.

nmorduch:
Deployed, and this appears to be fixed!

PR #1412, opened 2019-10-04
Merge `staging` (with `1399-tweak-og-image-size`) into `master`

nmorduch:
This wasn't testable with https://developers.facebook.com/tools/debug/og/object because all pages have the og:url set to the equivalent production url, but it does do what it is intended to do, so I'm merging it. I won't mark #1399  as closed until it's tested on production with a large image, though.


nmorduch:
Will deploy at the end of the day.

PR #1419, opened 2019-10-17
#1408 pdf generation to production

nmorduch:


nmorduch:
This works fine on staging! @kaedroho is there anything we need to do when deploying to production besides just the normal deployment pushing to heroku? And when this is in production will our heroku cost troubles be done?

kaedroho:
We would just need to copy the `PDF_GENERATOR_URL` setting from staging and this should all work!

Issue #1417, opened 2019-10-08
Draft pages are visible in some parts of the site


nmorduch:
The problem appears to be that when a page should not appear, whether that's because it's a draft or private or both, it can still show up on other pages that use React to grab a list of other pages. I'm guessing that this is an API setup problem.

- [x] Draft report showing up in a public RSS feed (#1403)
- [x] Draft edition of The Weekly showing up at [newamerica.org/weekly](https://newamerica.org/weekly)

AnnaGrieve:
Assigning to @jacobtoppm

jacobtoppm:
I've submitted a PR for review here: https://github.com/newamericafoundation/newamerica-cms/pull/1423

This should address the issue: Weekly queries were missing a `.live().public()` and Feed queries a `.public()`. 

I haven't changed how the latest edition is selected, aside from making it reject draft or private versions, as I didn't want to break any behaviour you're relying on. However, I've noticed that it doesn't have any ordering actually set in the query. While Wagtail's default ordering should grab the latest edition most of the time assuming you're adding them sequentially, this could potentially lead to problems if you update or re-publish an existing edition. Let me know if you would like me to add an ordering for you - either by the publication date or the edition number in the title.

nmorduch:
Thank you!

Let's make the ordering a separate ticket, since it's lower priority and I'd like to get this fix in sooner.

AnnaGrieve:
Assigning to @kaedroho to review PR

nmorduch:
Currently testing on staging!

nmorduch:
Looks good on staging!

nmorduch:
Closed with #1429

PR #1423, opened 2019-10-28
WIP: Draft pages fix

jacobtoppm:
Fix for #1417.

- Prevents private pages showing in feed
- Prevents draft and private pages showing in Weekly querysets or as the latest addition

### DIFF
--- a/newamericadotorg/api/weekly/views.py
+++ b/newamericadotorg/api/weekly/views.py
@@ -17,7 +17,7 @@ class WeeklyList(ListAPIView):
         return sorted(queryset, key=lambda edition: get_edition_number(edition));
 
 class WeeklyDetail(RetrieveAPIView):
-    queryset = WeeklyEdition.objects.all()
+    queryset = WeeklyEdition.objects.live().public()
     serializer_class = WeeklyEditionSerializer
 
 def get_edition_number(edition):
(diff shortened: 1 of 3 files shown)


nmorduch:
Thanks @jacobtoppm! @AnnaGrieve can we get review from a tbx BE dev?

btw jacob, I've written up a thing about the branching model we're using: https://github.com/newamericafoundation/newamerica-cms/wiki/Git-branch-setup

jacobtoppm:
> Thanks @jacobtoppm! @AnnaGrieve can we get review from a tbx BE dev?
> 
> btw jacob, I've written up a thing about the branching model we're using: https://github.com/newamericafoundation/newamerica-cms/wiki/Git-branch-setup

Ah, thanks for the link! I'll make sure to name branches according to that from now on.

newamericadotorg/api/weekly/views.py
```diff
@@ -17,7 +17,7 @@ def get_queryset(self):
         return sorted(queryset, key=lambda edition: get_edition_number(edition));
 
 class WeeklyDetail(RetrieveAPIView):
-    queryset = WeeklyEdition.objects.all()
+    queryset = WeeklyEdition.objects.live().public()
```
kaedroho: The `.public()` part performs a query on the `PageViewRestriction` model. This is causing a crash when this file is loaded as the database isn't ready at that time.

Best to wrap this in a `get_queryset` method so that it is run when the queryset is needed:

```python
def get_queryset(self):
    return WeeklyEdition.objects.live().public()
```

PR #1425, opened 2019-10-31
Vagrant

kaedroho:
This pull request adds a Vagrant file to make setup a bit quicker.

### DIFF
--- a/vagrant/provision.sh
+++ b/vagrant/provision.sh
@@ -1,0 +1,81 @@
+#!/bin/sh
+set -xe
+
+PROJECT_NAME=$1
+
+PROJECT_DIR=/vagrant
+VIRTUALENV_DIR=/home/vagrant/.virtualenvs/$PROJECT_NAME
+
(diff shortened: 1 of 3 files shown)


nmorduch:
What's the best way to go about reviewing this? Have Jacob review it? Test it?

vagrant/provision.sh
```diff
@@ -0,0 +1,81 @@
...
+alias djrun="dj runserver 0.0.0.0:8000"
+alias djrunp="dj runserver_plus 0.0.0.0:8000"
+
+source $VIRTUALENV_DIR/bin/activate
+export PS1="[$PROJECT_NAME \W]\\$ "
+cd $PROJECT_DIR
+EOF
+
+# Install node.js and npm
+curl -sSL https://deb.nodesource.com/setup_12.x | bash -
```
jacobtoppm: I think this needs to be 8 for building FE

kaedroho: I've removed it. Node should be used on the host machine anyway as it's really slow in Vagrant.

Issue #1426, opened 2019-10-31
Improve search algorithm


nmorduch:
Split from #1414, especially https://github.com/newamericafoundation/newamerica-cms/issues/1414#issuecomment-546925768

- [x] **Disable `partial_match`**
  This is a feature of Wagtail that allows any word in the query to match on the prefix of any word in the title. It is on by default in Wagtail but only makes sense for autocomplete searches. We should switch it off for actual searches though.
 

- [x] **Take document age into account**
  If the site uses Elasticsearch, we can customise the query so that older results have their score lowered to give preference to newer results.

AnnaGrieve:
Jacob to review

AnnaGrieve:
Change boosting on title in ranking - see if this works

nmorduch:
Opened a new PR. @kaedroho can you take a look at it and see what you think?

Can we define some page types as not showing up in search at all, and can we boost page types?
The search exposes some pages that shouldn't show up ever, e.g. https://www.newamerica.org/new-america-subscription-lists/education-policy-newsletters-events/

nmorduch:
Removing the estimate since it's old and I'm not even sure how much more we want to change this.

kaedroho:
> Can we define some page types as not showing up in search at all, and can we boost page types?

Yep, we can completely remove page types. Boosting is a bit trickier though since the title field is shared across all page types and there's not a way to boost individual documents. But we could probably do it as part of the query.

nmorduch:
Going to wait on #1476

PR #1430, opened 2019-11-01
[1426] Search algorithm changes

kaedroho:


nmorduch:
Is this ready for review? @kaedroho or @AnnaGrieve can you assign a Python reviewer

newamericadotorg/api/search/tests.py
```diff
@@ -13,28 +16,55 @@ def setUpTestData(cls):
...
 
     def test_search_all(self):
         url = '/api/search/?query=unique%20query'
         result = self.client.get(url).json()
 
-        self.assertEquals(getattr(result, 'error', False), False)
+        self.assertNotIn('error', result)
         self.assertEquals(result['count'], 50)
 
+    def test_partial_match_disabled(self):
```
jacobtoppm: These all look good, but maybe worth testing the date ordering as well? Ie that an older post is deprioritised?

search/backend.py
```diff
@@ -0,0 +1,71 @@
...
+                        'mustNot': filter_out
+                    }
+                }
+
+            return filter_out
+
+    def get_inner_query(self):
+        query = super().get_inner_query()
+
+        first_published_at_field_name = self.mapping.get_field_column_name(FilterField('first_published_at'))
```
nmorduch: And should we wait on #1456 so we can use the new date ordering field, or should we merge this, then update my branch and change it there?

nmorduch:
Addresses #1426 (comment is for linking convenience)

search/backend.py
jacobtoppm: It could use this or Date, but the new field is just for the cursor Pagination, so I think it's okay.

nmorduch: Oh right, it doesn't have to be unique here. Let's use `date` since that's the actual date of the page (often we use published + private instead of drafts, so the publish date is often wildly incorrect)

nmorduch: hm it occurs to me that non-posts should probably actually be high in the results, so using publication date as a fallback would have a downside of doing the opposite of that. e.g. if you search `education`, the (program homepage for education policy](https://www.newamerica.org/education-policy/) should come up at the top, and it's one of the earliest-published pages.

jacobtoppm: If you see my review below, this is what we intend to do following a discussion earlier - date will be used if available, and if date is not available, it will use today, ensuring more structural (ie non-Post) Pages without a date field go to the top of the ranking

kaedroho:
In testing, I noticed the Root page is in search results too, so dealing with that as well: https://www.newamerica.org/search/?query=root

kaedroho:
To add the date fallback, I've had to use `script_score`. This is sometimes disabled for security reasons [but there's no indication that SearchBox has done this](https://devcenter.heroku.com/articles/searchbox#api-limitations).

I'm following the workaround in https://github.com/elastic/elasticsearch/issues/18892

nmorduch:
re `script_score`:
- What is the security reason for which it is sometimes disabled?
- Why not use the syntax in the later comment? It looks like it was indeed added, though I may also not understand the difference well enough to know whether it matters which you choose. https://github.com/elastic/elasticsearch/issues/18892#issuecomment-432365923

AnnaGrieve:
According to @kaedroho - the syntax is due to the version of elastic search we are running. @kaedroho - can you add some more detail here?

nmorduch:
Makes sense re version! I'm interested to hear about the security reason for it being disabled sometimes, and wondering what's going on with tests.

kaedroho:
> Makes sense re version! I'm interested to hear about the security reason for it being disabled sometimes, and wondering what's going on with tests.

I'm not sure what the reason is. It might be because Elasticsearch has no authentication so could be a security issue if you haven't firewalled your Elasticsearch cluster properly.

Since we're using it very minimally (only returning the number `1`!), I don't think there would be any additional security concerns to worry about if SearchBox has already enabled scripting. There doesn't seem to be any docs on that or what scripting languages they support, so we'll have to try it on staging to find out I think.

nmorduch:
Thanks, @kaedroho! Does this need rereview from @jacobtoppm?

PR #1432, opened 2019-11-05
Download only original images

nmorduch:
Considering that `original_images` is 14+ GB and `images` is 9+ GB, it doesn't seem worth it to me to have the `downloadlocalimages` command download both.

I've made a minimal change to delete renditions of each image instead of downloading them. I would be interested to know if there's a better or more elegant way to do this, or if there's a reason it's not the best plan! I know this is a pretty trivial thing to change, but I think every time I touch Python is a good time for me to learn more about best practices.

Some questions:
- Should I also remove this line? It doesn't seem necessary since we're not putting anything there now.
  ```os.makedirs('media/images', exist_ok=True)```
- Would it make sense to just remove the print statements about deleting the renditions?
- Is the try/except for the renditions moot now?
- Is there a more efficient way to delete the renditions? I found that @gasman suggested [here](https://stackoverflow.com/a/42373300) that someone run `DELETE FROM wagtailimages_rendition;` but that did not work for me (and I'm not sure whether it could be done from within this command).

### DIFF
--- a/home/management/commands/downloadlocalimages.py
+++ b/home/management/commands/downloadlocalimages.py
@@ -30,7 +30,7 @@ class Command(BaseCommand):
 
             for r in i.renditions.all():
                 try:
-                    print('downloading ' + r.file.name + '...')
-                    s3.Bucket(BUCKET_NAME).download_file(r.file.name, 'media/' + r.file.name)
+                    print('deleting rendition ' + r.file.name + '...')
+                    r.delete()
                 except botocore.exceptions.ClientError as e:
                     print('The object: ' + r.file.name + ' does not exist')


kaedroho:
Just writing an update so this gets sent through to our internal kanban. Please ignore

kaedroho:
> Should I also remove this line? It doesn't seem necessary since we're not putting anything there now.

Yep, I think should do that. Django will create this folder automatically when it starts generating renditions, and it's not required for this task anymore.

> Would it make sense to just remove the print statements about deleting the renditions?

If we do this, we might as well change that whole loop to `image.renditions.all().delete()`. This should give a little speed improvement.

> Is the try/except for the renditions moot now?

Yes it is. It's only checking for a boto error, but boto is no longer called in the `try` block.

> Is there a more efficient way to delete the renditions? I found that @gasman suggested here that someone run DELETE FROM wagtailimages_rendition; but that did not work for me (and I'm not sure whether it could be done from within this command).

This is usually what we do at Torchbox because we usually fetch images with a `Fabric` task that doesn't have the Django ORM available.

I think replacing the loop with `image.renditions.all().delete()` will give the exact same result here (but it will be a bit slower as Django will check the filesystem for each file after deleting each rendition).

nmorduch:
thank you! this is v helpful review, v helpful to me learning python

nmorduch:
Obviated by #1489

PR #1433, opened 2019-11-07
Remove the "Share Our Space" page

nmorduch:
We're removing the "Share Our Space" page from the site, and it's hard coded into the menus as well as having React stuff specific to that page. I'm removing both the references to it and the custom layout, in part because we want to stop having pages set up like that anyway.

I removed everything that I found when searching for `share our space`, `share-our-space`, and `shareourspace`. I looked through a search for `share` just in case, but didn't fine anything.

### To test
localhost:8000/share-our-space should not look like https://www.newamerica.org/share-our-space/


nmorduch:
Reminder to myself that when I deploy this I'll also want to unpublish that page.

Issue #1416, opened 2019-10-08
[2-3hrs] Subpages of private reports are accessible


nmorduch:
From @mariaelkin in https://github.com/newamericafoundation/newamerica-cms/issues/1403#issuecomment-539159642
>In reports, if you publish a report that is password-protected, only the landing page prompts a password. If someone has a direct link to one of the interior page of the report, they can access it without putting in a password.

I think this is likely caused by the way urls are handled (which I believe is an attempt to keep Django urls and React router from stepping on each other's toes). e.g. https://github.com/newamericafoundation/newamerica-cms/blob/staging/newamericadotorg/urls.py#L54

nmorduch:
Basically..... sections of reports not being Pages causes various interesting side effects and problems. An interesting thing that will probbbbbbbably never cause a problem is that the program name doesn't actually matter at all in the url for a section of a report, e.g.:

https://www.newamerica.org/abcdef/reports/apprenticeship-review/higher-education-community-colleges-and-degree-apprenticeship/

So if two different programs ever had reports with the same name, with sections with the same name.... it doesn't seem like a high priority thing to test it out it to see what would happen.

nmorduch:
Another one is that you can't add redirects from things that are sections of programs. Which is more annoying as we're working on #1422

nmorduch:
There must be a better way to handle this url situation. Do we make it so that sections of reports have Pages in Wagtail? I expect there would be a number of complicated things about that plan. Can/should we change to a hash router for the React sections of the site? Some other less-buggy way to have React handle some-but-not-all of the urls?

kaedroho:
If I've understood this correctly, I think we could solve this by using [RoutablePageMixin](http://docs.wagtail.io/en/v2.1.1/reference/contrib/routablepage.html) on the `Report` page type to make it also respond on URLs directly beneath it. This would mean that Wagtail's routing would be used to find the `Report` page so it can check if it is password protected and this would also mean that the previous components in the path would be checked too.

Does that sound right?

AnnaGrieve:
Hi @nmorduch - we reckon this would require about 2-3 hours effort to implement, just checking you can confirm Karl's comments above?

nmorduch:
At a quick glance, that makes sense!

PR #1441, opened 2019-11-12
[WAITING] Add redirect from bretton-woods-ii to digi

### LINKED ISSUE #1422: [1hr] Create redirects for program name change
This program www.newamerica.org/bretton-woods-ii/ is going to be changing its name soon and rather than migrate their content to an entirely new program page, we'd like to change their URL to www.newamerica.org/digi. As a result, all links with /bretton-woods-ii will need to be redirected /digi.

I'm waiting on the official word on the timeline, but the deadline when they will debut the new name is likely going to be 11/13.

Is it possible to estimate your timeline for this?

nmorduch:
Fixes #1422 

Modeled off of https://github.com/newamericafoundation/newamerica-cms/commit/4fb6f34df2f5fca6346d0b55983da33c5239dfce

This solves a problem where links like https://staging.newamerica.org/bretton-woods-ii/blockchain-trust-accelerator/reports/blockchain-impact-ledger/blockchain-impact-ledger/ are unaffected by Karl's redirect solution (#1440). I noted in #1416 that the program name doesn't matter at all in the url for sections of reports... and redirects from those urls appear to do nothing at all, I assume because of the order in which things are in charge of routing. 

This is inelegant but I suspect it is the correct solution for tomorrow. Thoughts, @kaedroho? Thoughts on whether we should do this in concert with your solution or on its own? The downside of doing only this option is that we have to time it with the page slug change. It occurs to me that if we do both of them we would do your solution and then deploy the urls change, which would avoid timing problems related to changing the program page at the same time as deployment. On the other side, having a whole pile of redirects that we don't need seems cluttery, though I don't know whether that's a downside worth considering especially when we already have so many redirects that what does adding another hundred or two matter.

### DIFF
--- a/newamericadotorg/redirects.py
+++ b/newamericadotorg/redirects.py
@@ -10,3 +10,9 @@ def dual_language_learners(request, **kwargs):
     path[2] = 'english-learners'
     url = '/'.join(path)
     return redirect(url)
+
+def digi(request, **kwargs):
+    path = request.path.split('/')
+    path[1] = 'digi'
+    url = '/'.join(path)
+    return redirect(url)
--- a/newamericadotorg/urls.py
+++ b/newamericadotorg/urls.py
@@ -47,6 +47,7 @@ urlpatterns = [
 
     url(r'^international-security/future-property-rights/[^.]*$', redirects.future_property_rights),
     url(r'^education-policy/dual-language-learners/[^.]*$', redirects.dual_language_learners),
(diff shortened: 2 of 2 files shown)


nmorduch:
Oh right, the redirects rfc! I think that will be A Good Change.

I'm going put your script (#1440) into production too so that the timing thing isn't such a big deal, since that one I can do ahead of when they're ready for this. And I'm going to wait on merging this one into staging because we don't want it to accidentally get onto production.

PR #1443, opened 2019-11-18
Copy some timeline legend styles from one place to another

nmorduch:
Hopefully this will encourage the timeline to look better


nmorduch:
Fun fact: the timeline doesn't work at all for me locally but so I don't actually know what this will look like, and am just checking it out on staging 🙃

Issue #1453, opened 2019-12-05
[2-3hrs] Some items duplicated on some program homepages


nmorduch:
See https://www.newamerica.org/education-policy/

It seems that only the first few are duplicated, and I really have no idea why. They're not doubled in the backend as far as I can tell.

nmorduch:
We've now had 3 independent complaints about this, which suggests to me that it's recent, which suggests to me that it's related to the recent changes

kaedroho:
This is likely (but not 100% sure) to have been caused by the React changes I made to allow the new pagination method to be used on the publications view. I think it would take around 2 - 3 hours to investigate and resolve.

nmorduch:
We also got this follow-up:
> some people are reporting that they’re having a similar problem with the displays on the publications page when they choose “show more”. As in the first few blogs that display are repeating. (And all the X’s have turned into slashes when you open/close the drop downs).

kaedroho:
Spent some time looking at this today, I think the issue exists somewhere within the `InfiniteLoadMore` component. I haven't pinned down the issue with it yet though.

The API isn't returning duplicates, and it looks like the reducer is combining them correctly too.

I've opened up a WIP pull request that will later contain the fix for this (currently contains a little bit of refactoring of `InfiniteLoadMore`). https://github.com/newamericafoundation/newamerica-cms/pull/1454

nmorduch:
This is looking good to me on staging!

PR #1454, opened 2019-12-13
[1453] Filter results on program page to not include features

kaedroho:
See #1453 

Features were being displayed twice, both at the top and within the
regular results.

This filters the results so that features are not repeated.


nmorduch:
This one too is ready for staging if you think my review is sufficient. But let's only put one on staging at a time

Issue #1459, opened 2019-12-19
"Load more" button on publications page sometimes loads first page of content again


nmorduch:
gotta love an inconsistently-occurring problem...

when one experiences this (mostly has been happening only locally, but occasionally found in production), every time you hit load more, you just get that same page again

nmorduch:
I already wasn't getting this on staging, so I have no idea if it's fixed things! But the publications page seems to be working normally, so what more can we ask for

PR #1460, opened 2019-12-19
[1459] Disable browser API cache

kaedroho:
This removes the browser-side caching of API responses.

It's been causing issues in the publications index and program pages, we don't think this cache provides much performance benefit and removing it solves these issues and simplifies the code a bit.


nmorduch:
Addresses #1459

nmorduch:
If you think my review is sufficient, you can merge this one to staging

PR #1465, opened 2019-12-20
Use BEM for card types

nmorduch:
`.card.list.program` was getting styles for `.program` that it was not supposed to be getting.

e.g. the card for education policy here: https://staging.newamerica.org/search/?query=education+policy


nmorduch:
and the styles for books show up here https://staging.newamerica.org/books/

PR #1467, opened 2020-01-08
Update allowed_hosts

### LINKED ISSUE #1076: Set Allowed Hosts for Site
- Currently in mysite/settings allowed sites is set to *, which is a potential security risk because it would allow anyone to host our site - should change to na-staging.herokuapp and production/newamerica.org

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/113920320650537/152981350214784)

jckpn:
Addresses #1076

### DIFF
--- a/newamericadotorg/settings/production.py
+++ b/newamericadotorg/settings/production.py
@@ -16,7 +16,10 @@ SECRET_KEY = os.getenv("SECRET_KEY")
 #ADMINS = [('Kirk', 'jackson@newamerica.org'), ('Andrew', 'lomax@newamerica.org')]
 
 # Will be changed to final host url
-ALLOWED_HOSTS = ['*']
+ALLOWED_HOSTS = [
+    '.newamerica.org',
+    'na-staging.herokuapp.com'
+]
 
 SECURE_SSL_REDIRECT = True
 SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
--- a/newamericadotorg/settings/dev.py
+++ b/newamericadotorg/settings/dev.py
@@ -4,7 +4,7 @@ from .base import *
 # SECURITY WARNING: don't run with debug turned on in production!
 
 DEBUG = True
-ALLOWED_HOSTS = ['*']
+ALLOWED_HOSTS = []
 
 for template_engine in TEMPLATES:
     template_engine['OPTIONS']['debug'] = True


newamericadotorg/settings/production.py
```diff
@@ -15,8 +15,7 @@
 
 #ADMINS = [('Kirk', 'jackson@newamerica.org'), ('Andrew', 'lomax@newamerica.org')]
 
-# Will be changed to final host url
-ALLOWED_HOSTS = ['*']
+ALLOWED_HOSTS = ['www.newamerica.org']
```
nmorduch: In order for `staging.newamerica.org` also to work, do we want to use `*.newamerica.org` here?

nmorduch: Looks like my timing was off here!

jckpn: Good point, *.newamerica.org would probably be better than www.newamerica.org.

jckpn: Deleting the line in the dev settings may throw up errors as I think that's where `allowed_hosts` is defined, but I'll go ahead and make the array empty as the docs say, just to be safe.

jckpn:
Deleting the line in the dev settings may throw up errors as I think that's where allowed_hosts is defined, but I'll go ahead and make the array empty as the docs say, just to be safe.

newamericadotorg/settings/production.py
```diff
@@ -15,8 +15,10 @@
 
 #ADMINS = [('Kirk', 'jackson@newamerica.org'), ('Andrew', 'lomax@newamerica.org')]
 
-# Will be changed to final host url
-ALLOWED_HOSTS = ['*']
+ALLOWED_HOSTS = [
+    '.newamerica.org',
```
nmorduch: Will this work? Should it not be `*.newamerica.org`?

jckpn: Django uses a bit of a weird syntax for url wildcards. The docs says:

```
A value beginning with a period can be used as a subdomain wildcard: '.example.com' will match example.com, www.example.com, and any other subdomain of example.com.
```

nmorduch: Oh weird. Thanks!

nmorduch:
Local dev still works for me, not sure if there's a good way to test what this will do in production, and not confident that I have the expertise to assess the change. @kaedroho can you take a look?

Additionally, it would be great to squash the commits before merging this one.

newamericadotorg/settings/production.py
kaedroho: Nice, I didn't know this!

nmorduch:
Assigning to myself to update the branch and merge.

jckpn:
Sorry @nmorduch, I couldn't find a way to squash the commits without creating a new branch/pr altogether which I thought might mess with things -- just let me know if you still want me to :)

nmorduch:
Noting here that I directed Jack to https://github.com/k88hudson/git-flight-rules#i-need-to-combine-commits

Issue #1468, opened 2020-01-09
Can't save post under subprogram


nmorduch:
Getting this error:
>'Subprogram' object does not support indexing

Seems to be related to [this change](https://github.com/newamericafoundation/newamerica-cms/pull/1456/files#diff-0f16a55344e1d14f1f70280caeba85d9R652), because removing `.first()` from that line makes the error stop happening and pages can be saved. But I haven't looked into it deeply enough to know whether that's the correct answer.

Urgency: currently, new blog posts can be published, but existing blog posts cannot be editing

nmorduch:
More specifically, having this problem with the BLLx blog, haven't tested elsewhere https://www.newamerica.org/better-life-lab/better-life-lab-experiments/bllx-blogs/experiment-no-13-smart-resolutions/

PR #1469, opened 2020-01-09
Fix/1468 save post under subprogram

nmorduch:
Addresses #1468 

Not sure enough about what's going on here to know whether this is the best way to test nor whether this is the correct solution, but it _is_ working for me locally.

### DIFF
--- a/home/models.py
+++ b/home/models.py
@@ -652,7 +652,7 @@ class Post(Page):
             subprogram = Subprogram.objects.ancestor_of(self).first()
             if subprogram:
                 relationship, created = PostSubprogramRelationship.objects.get_or_create(
-                    subprogram=subprogram[0].specific,
+                    subprogram=subprogram.specific,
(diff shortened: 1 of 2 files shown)


home/models.py
```diff
@@ -649,7 +649,7 @@ def save(self, *args, **kwargs):
             if created:
                 relationship.save()
 
-            subprogram = Subprogram.objects.ancestor_of(self).first()
+            subprogram = Subprogram.objects.ancestor_of(self)
```
kaedroho: I think we should keep `.first()` here and instead remove the `[0]` from `subprogram[0].specific` below. This way we will only fetch a single subprogram

nmorduch: Ohh I understand what the problem was. That either we should fetch a list/array/whatever then take the thing at index 0, or we should fetch just the first, and then there's nothing to index since it's one object.

PR #1470, opened 2020-01-09
Add migration for fellowship year

nmorduch:
It's not a long-term fix, but.


nmorduch:
Added an issue for fixing properly (#1471) and added a mention in the wiki

Issue #1397, opened 2019-08-19
"Former fellows" button doesn't go anywhere


nmorduch:
The button at the bottom of this page is supposed to load the former fellows.
https://www.newamerica.org/fellows/our-people/

The page they're using instead is
https://www.newamerica.org/fellows/about/alumni-roster/

This could be solved either by a) making it work correctly, with clicking the button loading all of the former fellows or by b) making that button link to the alumni roster page. @mariaelkin, thoughts? I lean toward the latter, since they have content on the alumni roster page.

kevinhowbrook:
Hey @nmorduch 

Having looked into this, it looks like the button is working but the request the button calls is responding with no data:
`https://www.newamerica.org/api/fellow/?program_id=132&page_size=25&former=true`

Fixing the button is probably the best option here, or more specifically pinning down why it's no longer working. I suspect it's because some people that are `former=True` are not linked to any programs, e.g https://www.newamerica.org/admin/pages/12398/edit/

To tackle this it's best to relate some `persons` pages to a program and set them as `former=True` on staging.

Option b isn't a bad solution, but if there is functionality like this built-in it would be better to fix and use it, however if over the life cycle of the site that the way former fellows are handled (ie being unrelated to programs once set to former) then it's wise to replace the button with your suggestion.

nmorduch:
Let's just make the button link to the page they're using, at least for now. Especially since I should eventually be changing how this page works.

mariaelkin:
Yes, let's just change the button to go the link that they're using, for the time being. The Fellows Program is the only one that actively uses the Former Fellows button. Fixing former fellows' bios so they have a designated program in addition to being marked former is something we can do, but will take some time. @kevinhowbrook is it easy for you to pull a list of people who are causing the problem? If not, nbd, we can just cross-reference from the alumni roster page. This is a medium-low priority.

nmorduch:
@mariaelkin do I understand correctly that making the button go to the alumni roster is high priority, and having it be functional as intended is what is medium-low priority?

mariaelkin:
What's the time estimate on just making the button go to the alumni roster? Depending on time, I think we make that fix and call it a day, but it's the least high priority of the high priority things.

Thinking about it more, I don't think it's worth our time to make it work as intended. Even working as intended, it wasn't doing what the program needed.

nmorduch:
[...]

PR #1472, opened 2020-01-09
For fellows program, hard code link to former fellows page

nmorduch:
Addresses #1397 

Not a nice solution but the quickest acceptable one.


nmorduch:
Good catch on the class, I def didn't even read its name. I'm going to leave it incorrect though because I don't think it's worth putting much effort into this right now.

Issue #1476, opened 2020-02-04
[1-3 hrs] Update Wagtail to 2.8


nmorduch:
Split from #1365

AnnaGrieve:
@nmorduch to confirm preferred approach for testing upgrade - use existing Stage site or set up new stage - either manually or in Heroku review app (noting database approach - don't want live data in Heroku review app?)

PR #1490, opened 2020-02-24
[1476] Wagtail update

kaedroho:
Addresses #1476 

This pull request updates Django to 3.0, Wagtail to 2.8 and all other dependencies to their latest versions.


nmorduch:
Added another. can't edit blog posts

kaedroho:
>  Fonts aren't loading

The fonts don't seem to be checked into the repo so they were probably uploaded to S3 directly (and `na-wagtail-update` isn't using S3 for static). Do you know why this was? Maybe a licencing issue?

>  Will using logdna only in production mean it's used on staging? I think no since I think staging uses production settings?

We reuse the same LogDNA instance for production/staging/na-wagtail-update. It's all configured on Heroku (just takes stdout).

>  Do we need to update allowed hosts for the other server?

Done this on this branch: https://github.com/newamericafoundation/newamerica-cms/tree/wagtail-update-staging

>  Search is broken, FE and BE

Sorry, I provisioned the wrong Elasticsearch version. This should be fixed now!

kaedroho:
>  Can't add or edit some page types.

These should be fixed now by the `Fix rendering of LocationWidget` commit.

kaedroho:
The card issue should be fixed now, it was caused by calling `page.content_type` directly but the output of that has changed in a recent Django release. I've updated all other occurrences of this to be more explicit about the representation of the model we want.

nmorduch:
## Current list

- [ ] Let's get this server set up like staging sooner rather than later so that it will have fonts/icons, and that way we'll know nothing's wrong with those.

From Joanne's testing:

- [ ] In photo and document uploads, it’s asking for the “Collection” and options are “Root” and “test”
- [ ] Tried to add a new report page, filled out the info in the first page (so didn’t have a chance to upload a doc) and clicked “Save Draft” and got an internal server error
  - https://na-wagtail-update.herokuapp.com/admin/pages/add/report/report/20185/
- [ ] I am able to add a Draft/Unpublished page as a featured story on a program homepage and it shows up (you get a “Page not found” when you click on it though) [Naomi: this might not be a new bug]
- [ ] .gif upload took a while for a single gif [Naomi: is this just because of the server setup?)
- [ ] Previewing the page with the gif got an error message (application error from heroku). It worked the second time I tried to preview.
- [ ] Uploaded 10 gifs at once using bulk uploader, they seemed to successfully upload one by one, but now it won’t load the image folder or the image selector. Heroku times out and gives same error message as above

newamericadotorg/wagtailadmin/widgets.py
```diff
@@ -1,10 +1,15 @@
...
-from django.shortcuts import render
 from django.template import loader
 
+
 class LocationWidget(widgets.TextInput):
     template_name = 'wagtailadmin/location.html'
 
-    def render(self, name, value, attrs=None):
-        attrs['value'] = value;
+    def render(self, name, value, attrs=None, renderer=None):
```
jacobtoppm: What does `renderer` do here? It looks unused

kaedroho: Django passes it through and would crash if we don't have the kwarg. We don't use it because we always render with a template.

jacobtoppm: Oh I see - makes sense, thanks.

nmorduch:
Added above, a definitely blocking problem: document upload -> Wagtail sections is working, but Wagtail sections -> PDF is not

kaedroho:
Hi Naomi,

>  This page isn't showing anything, looks like a JS problem? https://na-develop.herokuapp.com/political-reform/reports/census-test-report/draft-graphics/

This seems to be broken on staging as well, so it's probably not related to the Wagtail update.

> Draftail has the wrong icon for block quotes (and is still showing an h5 in the place where I think we want it to show a p, but that's not new)

I've changed the icon back and moved the H5 to the end. There doesn't seem to be a rich text feature in Wagail that gives us a P button so I'll leave this for later.

ababic:
@kaedroho / @nmorduch Just wanted to highlight that this PR should be updated to use Wagtail 2.8.1 instead. And (after rebasing) should revert the changes in #1516, as that should no longer be necessary.

nmorduch:
To do for @kaedroho on Monday morning:
- [ ] Review changes I made today (which also are merged to `develop`, which is deployed to both the dev and staging servers). The changes are:
  - Made [a change](https://github.com/newamericafoundation/newamerica-cms/pull/1490/commits/055094361dbdac05bcd1db1c2f2531183afe4757) so that I would be able to choose the changes from this branch when resolving the conflicts introduced in #1515 without undoing the fix to the page type display
  - Merged in `master`
  - Merged in `static-serving` such that this obviates #1513 (because without those changes, this wasn't working on staging
  - Updated to Wagtail 2.8.1
- [ ] Merge this pr to `master` (which will trigger deployment to staging)
- [ ] Check that all is well on staging (not that it will actually be different than it is now)
- [ ] Deploy to production, but remove the extra variables first
  - [ ] Remove `STATIC_BUCKET_NAME` and `STATIC_URL` from the production settings (If you do this after deploying the changes, you'll have to deploy again in order to make it work, so you might as well do this first. It would be great to do this well before the business day starts here, since it breaks the site.)
  - [ ] Deploy `master` to production
- [ ] Check that all is well on production, including that the css is being served by Heroku and the extra static assets still from S3


And then maybe sometime we'll actually try that "promote to production" button in Heroku, since after this is all done we should have static set up in a way that will allow that!

home/templatetags/utilities.py
```diff
@@ -335,4 +335,7 @@ def model_display_name(page):
 	"""
 	Gets verbose name of the page's model
 	"""
-	return page._meta.verbose_name.title()
+	if page._meta.verbose_name.title() == 'Other Post':
```
kaedroho, line 338: I think it would be better to use `isinstance` here:

```python
if isinstance(page, OtherPost):
```

home/templatetags/utilities.py
```diff
@@ -335,4 +335,7 @@ def model_display_name(page):
...
+		return page.other_content_type
```
kaedroho, line 339: Looks like this file is indented with tabs instead of spaces!

Issue #1455, opened 2019-12-13
[5 hrs] Stop having editions in The Weekly | BE


nmorduch:
The Weekly is going to stop having editions and start posting things as they're ready. This obviates #1424. I think we'll want to use edition pages ([Edition.js](https://github.com/newamericafoundation/newamerica-cms/blob/staging/newamericadotorg/assets/react/weekly/components/Edition.js)) as the base for the new Weekly homepage.

I can definitely do the design changes, and am less confident when it comes to changing React/Python parts. Torchbox (probably @kaedroho), can you look at this plan and see if you think it makes sense?

- [x] Backend work (Tbx)
  - [x] Update the Weekly API so that it returns all articles rather than editions [weekly/serializers.py](https://github.com/newamericafoundation/newamerica-cms/blob/staging/newamericadotorg/api/weekly/serializers.py)
  - [x] Add pagination (ie the "load more" button)
  - [x] Move article pages all to be children of the weekly homepage and make redirects.
  - [x] [/weekly](https://www.newamerica.org/weekly) should point to the Weekly homepage, not to editions

FE (Naomi) split to #1486
- New Weekly homepage design
  - Like an edition page, including keeping a full-width article at the top
  - Move subscription CTA to a card (like on program homepages), put it in the second row. Maybe the first card in the second row. Or maybe 2 cols wide?
  - Add date to article cards
  - Probably remove black bar entirely
  - Add pagination (ie the "load more" button)
- Weekly article design
  - Black bar at the top should link back to the Weekly homepage, not to an edition
  - Take away the text that says "Weekly Article" under the headline

AnnaGrieve:
@kaedroho to take a look today

kaedroho:
This sounds good to me! Just a couple of comments:

We would also need to update the Weekly API so that it returns all articles rather than editions
Do we need to provide redirects for past editions? It might be difficult to support this if the pages have to be restructured.

nmorduch:
Would the redirects be easy if we use [that thing you wrote for making redirects from pages to themselves](https://github.com/newamericafoundation/newamerica-cms/blob/staging/home/management/commands/pregenerate_redirects.py)?

Will add the bit about the API. And then can you estimate the BE work?

kaedroho:
> Would the redirects be easy if we use that thing you wrote for making redirects from pages to themselves?

[...]

PR #1493, opened 2020-02-25
[1455, waiting on FE] Stop having editions in The Weekly (backend)

kaedroho:
Addresses #1455


weekly/models.py
```diff
@@ -14,56 +14,25 @@
...
-
-    def get_context(self, request):
-        context = super().get_context(request)
-        if getattr(request, 'is_preview', False):
-            import newamericadotorg.api.weekly
-            revision = PageRevision.objects.filter(page=self).last().as_page_object()
-            weekly_data = newamericadotorg.api.weekly.serializers.WeeklyEditionSerializer(revision, context={'is_preview': True}).data
-            context['initial_state'] = json.dumps(weekly_data)
+    parent_page_types = []
+    subpage_types = []
```
nmorduch: Why do we do this instead of actually deleting the model? So that it doesn't break before we run the migration command?

weekly/management/commands/remove_weekly_editions_migration.py
```diff
@@ -0,0 +1,58 @@
...
+        for article in WeeklyArticle.objects.all():
+            print("moving article", article)
+            Redirect.objects.create(
+                old_path=Redirect.normalise_path(article.relative_url(site)),
+                redirect_page=article,
+            )
+
+            # Find a slug that isn't taken under the index
+            old_slug = article.slug
+            article.slug = find_available_slug(weekly_index, article.slug)
```
nmorduch: Not sure this is actually working; neither http://localhost:8000/weekly/which-women-can-really-walk-out/ nor http://localhost:8000/weekly/which-women-can-really-walk-out-1/ is working for me

pulled from
```
moving article Which Women Can Really Walk Out?
changed article slug which-women-can-really-walk-out => which-women-can-really-walk-out-1
```

weekly/models.py
kaedroho: Yep, that's right. These will continue to exist until that migration is run. We will have to follow up with another PR to remove these after this is deployed and migration is run.

weekly/management/commands/remove_weekly_editions_migration.py
nmorduch: Oh actually most of them aren't working. maybe any that aren't the latest. i'll see if doing the FE fixes this.

nmorduch: ok, it was the FE that was the problem, AND i figured out why the renaming was needed on such specific names. it's because sometimes old articles have been reposted!

nmorduch:
This is just waiting on the FE!

PR #1494, opened 2020-02-25
[1426] search algo changes

nmorduch:
Addresses #1426. See original pr #1430 

Requires running `python manage.py update_index`

This still isn't amazing. Can we define some page types as not showing up in search at all, and can we boost page types?


nmorduch:
Got this error on the dev server when running update_index but that's about it being on the dev server, yeah?

``` 
elasticsearch.exceptions.TransportError: TransportError(500, 'generation_exception', 'reached maximum index count for current plan')
```

PR #1495, opened 2020-02-25
1478 full bleed images

nmorduch:
Addresses #1478 

This is mediocre but I don't know that it's worth doing it better or naming things better.

Checking the box for "theme_full_bleed" will make the image on a report full width.

And then you can make full bleed images at the top of report sections by adding an image with the full width size, hiding the section title, and adding a title in the paragraph below.

But this actually has already been tested on staging, so you can just do code review and not test (and there's a report set up for it on staging where you can see it once merged)


newamericadotorg/templates/blocks/image_block.html
```diff
@@ -28,13 +28,17 @@ <h4 class="inline margin-0">{{ value.figure_title }}</h4>
       <a href="{{ img_original.url }}" target='_blank' >
         {% if value.use_original %}
           <img src="{{img_original.url}}" class="original"/>
+        {% elif value.width == 'width-full'%}
+          <div class="figure__image__background-image" style="background-image: url({{ img.url }});"></div>
```
kaedroho: `img` is only 800px wide (https://github.com/newamericafoundation/newamerica-cms/pull/1495/files#diff-bc8a79da59971955acdd2da224d9dcc8L5) is this big enough for full width?

nmorduch: I think it's 1600 (https://github.com/newamericafoundation/newamerica-cms/blob/staging/newamericadotorg/api/report/serializers.py#L35) and it's not amazing but I'm going to call it good enough for now.

Issue #1475, opened 2020-01-16
[1 hr] Weekly articles ordered incorrectly in Publications


nmorduch:
https://www.newamerica.org/publications/?after=2020-01-05&before=2020-01-16
https://www.newamerica.org/weekly-articles/?after=2019-12-01&before=2020-01-16

It's just the weekly articles from January, and they're getting sorted definitely after December ones, possibly after all publications

kaedroho:
Looks like the issue is that weekly articles are bypassing the `Post.save()` method: https://github.com/newamericafoundation/newamerica-cms/blob/staging/weekly/models.py#L61

This means that the `ordered_date_string` field isn't being set. I'm not sure why this is being bypassed. All the logic except for setting `ordered_date_string` would be bypassed by Weekly articles.

So the solution would be to remove these two lines and run a script to populate this field for weekly articles.

nmorduch:
Thanks! I can take this one then :)

PR #1496, opened 2020-02-25
1475 weekly article ordering

nmorduch:
Addresses #1475 

Maybe this migration was a silly way to do this? But it does do the thing. If it was indeed a silly way to do this, what would be the more normal way to do it, and is it worth changing to that?


kaedroho:
Nope, this isn't silly at all! Looks good to me

I wonder if we should remove the `blank=True` from the field on the model so Wagtail will raise an error if it happens again?

PR #1498, opened 2020-02-26
Sentry configuration

kaedroho:


newamericadotorg/settings/production.py
```diff
@@ -73,6 +73,22 @@
...
+    import sentry_sdk
+    from sentry_sdk.integrations.django import DjangoIntegration
+
+    sentry_sdk.init(
+        dsn=os.environ['SENTRY_DSN'],
+        integrations=[DjangoIntegration()],
+
+        # If you wish to associate users to errors (assuming you are using
+        # django.contrib.auth) you may enable sending PII data.
+        send_default_pii=True
```
jacobtoppm: Do we know what form this data's in? I'm assuming it's not pasting user emails, etc into the log?

kaedroho: Yep, it will be pasting user names/emails into the log. But the only users are Wagtail editors so I think this would be OK

nmorduch: Yeah, I'm ok with collecting usernames/emails of wagtail editors, and things like ip addresses we're already collecting with google analytics

Issue #1486, opened 2020-02-20
[3 hrs] Stop having editions in The Weekly | FE


AnnaGrieve:
Split from #1455

- [x] New Weekly homepage design (Naomi)
  - [x] Like an edition page, including keeping a full-width article at the top
  - [x] Move subscription CTA to a card (like on program homepages), put it in the second row. Maybe the first card in the second row. Or maybe 2 cols wide?
  - [x] Add date to article cards
  - [x] Probably remove black bar entirely
  - [x] Add pagination (ie the "load more" button)
- [x] Weekly article design (Naomi)
  - [x] Black bar at the top should link back to the Weekly homepage, not to an edition
  - ~Take away the text that says "Weekly Article" under the headline~

Backend work (Tbx) in #1455
 - Update the Weekly API so that it returns all articles rather than editions [weekly/serializers.py](https://github.com/newamericafoundation/newamerica-cms/blob/staging/newamericadotorg/api/weekly/serializers.py)
- Add pagination (ie the "load more" button)
- Move article pages all to be children of the weekly homepage and make redirects.
- [/weekly](https://www.newamerica.org/weekly) should point to the Weekly homepage, not to editions

nmorduch:
Blocked by #1455

nmorduch:
This appears to be working only in Firefox on staging. Need help troubleshooting

kaedroho:
@nmorduch Which version of Chrome did you test with? It seems to work for me on Chrome 80.

nmorduch:
It appears that it works for me in Chrome at work, but at home (where I had to do the basic auth) it did not

nmorduch:
I'm pretty convinced that it was a weird thing with the auth, especially since that's happened to me before with things like the publications page

PR #1502, opened 2020-02-27
1486 stop having editions in the weekly (frontend)

nmorduch:
Supersedes #1493 
Addresses #1486 

I did a really messy job of the react on the weekly homepage but every time I tried to clean it up the whole thing broke, so I decided I'm just not there yet and left it


weekly/templates/weekly/weekly_article.html
```diff
@@ -1 +1,15 @@
-{% extends "./weekly.html" %}
+{% extends "post_page.html" %}
+
+{% block breadcrumbs %}
+  <div class="weekly-header">
+    <div class="container">
+      <h5 class="weekly-header__button button--text with-caret--left white margin-0">
+        {% if self.get_ancestors|length > 2 %}
```
kaedroho: Could we make this more generic?

Maybe something on the lines of:

```HTML+Django
{% for ancestor in self.get_ancestors|slice:"2:" %}
    <a href="{{ ancestor.url }}"><u>{{ ancestor.title }}</u></a>
{% endfor %}
```

nmorduch: I want to leave it for now since it's the same as the existing breadcrumbs, but yes, later, something like that.

PR #1504, opened 2020-02-28
Optimise populate_post_ordered_date_string migration

kaedroho:
Found a way to run this migration with a single query, which takes the run time down from a few minutes to a few seconds.


kaedroho:
Sorry, made PR from wrong branch!

PR #1506, opened 2020-02-28
[1483] New search component with page-based pagination

kaedroho:
This reimplements the search component to switch to page-based pagination (away from load-more). It's reimplemented this way to avoid having to do much work with the existing Redux infrastructure.

Currently, it's using `/?page=<id>` instead of the `/?after=`/`/?before=` that we discussed because items are ordered by score which isn't easy to filter by in Wagtail at the moment.

This component actually works as-is with the `/api/post/` endpoint (even though the pagination works differently!), so it shouldn't be hard to refactor it for the Publications page once it supports filtering.


package.json
```diff
@@ -11,7 +11,7 @@
     "start": "source .env && source venv/bin/activate && ./manage.py runserver",
     "start:lite": "webpack-dev-server --config ./webpack.config.lite.js --mode development",
     "build": "NODE_ENV=development webpack",
-    "dev": "source .env && NODE_ENV=development webpack --watch",
```
kaedroho: Accidentally committed this.

nmorduch: Are you doing something other than this?

kaedroho: It crashes on my computer because my version of node doesn't run scripts with `bash`, it runs them with `sh`

nmorduch: Ok to request putting it back in a future commit?

nmorduch:
I'm going to try updating npm when testing locally

nmorduch:
Karl also approves. I want to finish #1511 before merging this.

nmorduch:
Rebased off master, and deploying to the dev server.

PR #1508, opened 2020-03-04
1481 Add verbose names for a few more page types

nmorduch:


home/models.py
```diff
@@ -259,6 +259,9 @@ def get_context(self, request):
 
         return context
 
+    class Meta:
+        verbose_name = 'Homepage'
```
kaedroho: Verbose names should be lowercase as Django will capitalise them where needed (unless it's a name/acrynym/etc)

nmorduch: Except that we use them in the FE so that doesn't work for us right now 🙃

nmorduch:
Karl says this is ok to merge

PR #1513, opened 2020-03-17
Add EXTRA_STATIC_URL

kaedroho:
`STATIC_URL` will now point at Django-managed static, but we are going to leave fonts, logos, social images, etc in the S3 bucket. So I've added `EXTRA_STATIC_URL` which will point at the S3 bucket until we find a new place for these files.

To deploy:

 - Set `EXTRA_STATIC_URL` to the same value as `STATIC_URL` (must be done before deployment as this value is read during the build stage)
 - Deploy this branch
 - Remove `STATIC_BUCKET_NAME` and `STATIC_URL`


nmorduch:
So this is the #1511 plus one commit, yeah? So it supersedes that pr?

kaedroho:
We need that PR as it updates Whitenoise. It's based on it but doesn't supersede it, we should merge the Wagtail update first then this.

nmorduch:
What I want to do is set the value of `EXTRA_STATIC_URL` as what `STATIC_URL` currently is on production, for all three servers, right?

kaedroho:
> What I want to do is set the value of EXTRA_STATIC_URL as what STATIC_URL currently is on production, for all three servers, right?

Yep, that should work!

nmorduch:
If you look on the dev server, it's currently looking for images at STATIC_URL when it needs to be at that plus /static

nmorduch:
And then it works fine when I changed EXTRA_STATIC_URL to account for that, but it would probably be better to add `/static` in the code

PR #1516, opened 2020-04-14
[NOTICKET] Wagtail security patch

ababic:
Temporarily disable revision comparison as per security guidance on
https://github.com/wagtail/wagtail/security/advisories/GHSA-v2wc-pfq2-5cm6


ababic:
Hi @nmorduch. I've been asked to patch this project as part of Torchbox's response to the recently uncovered security vulnerability in Wagtail. If you could review and possibly let me know if/when it can be deployed, that would be fabulous. Thank you :)

nmorduch:
@ababic forgot to tell you this was deployed!

Issue #1533, opened 2020-06-10
Performance issues


nmorduch:
- Slow loading times
  - weekly page took 36 seconds for everything below the main toolbar to load whereas the main Publications page only took 5 seconds
  - The content grid on the Event page took 17 seconds
- Memory overload for web dynos
  - I just decreased to 5 workers but not sure this will help
- System exits in Sentry might be related?

kaedroho:
Naomi and I had a look through New Relic during our catchup today:

 - The Weekly takes an average of 7 seconds to load. A lot of time is spent in Python and template rendering. It runs about 150 queries for images so I think it's most likely to be image processing causing the issue. We don't think this view should be fetching any images though because the listing is rendered with React so this might be an easy fix
 - The Meta API is called on every single request and takes over a second each time to respond. If this is returning the same response for every user we can cache this which should make reduce the response time to a few milliseconds
 - The system exits are because the Gunicorn workers are timing out. I think there's a chance this might be reduced/eliminated when we fix the meta API (which currently uses up around 50% of total request time on its own).

kaedroho:
We had another look at this. There's also a performance issue with editing reports. Sometimes the report view even times out.

According to New Relic, It's spending a lot of time rendering streamfield blocks. Switching to [wagtail-react-streamfield](https://github.com/wagtail/wagtail-react-streamfield/) might solve this.

nmorduch:
Search still has problems, probably fixed in #1494. Post api is still having problems.

nmorduch:
Changed to medium priority because we've made some progress on this but there's still more we could do. ie it's still bad but not abominable.

nmorduch:
This is not worth it in the near future if we're redoing our FE

PR #1536, opened 2020-08-12
Performance improvements

jacobtoppm:
For #1533 

- Cache meta API, which is run on every request and is identical
- Switch to using the Wagtail dynamic image serve view, as rendition links are often serialised and then not used
- Optimise the Weekly view: 
    - Add some prefetching/select_related
    - Split up serialisers, and don't serialise big unnecessary fields for listings where they're not used (particularly rendering the post and body)


home/templatetags/utilities.py
```diff
@@ -71,7 +71,7 @@ def pluralize(num_items, label):
 # generates byline for all post types - calls byline prefix tag to get apporpriate prefix
 @register.simple_tag()
 def generate_byline(post_type, authors):
-	authors = authors.order_by('pk')
+	authors = authors.select_related('author')
```
kaedroho, line 74: Definitely nicer this way as this `order_by('pk')` is used in a few places, but I'm not sure if it's a performance improvement since it's just the default value for `.order_by`.

newamericadotorg/api/author/serializers.py
```diff
@@ -2,8 +2,9 @@
 
 from rest_framework.serializers import ModelSerializer, SerializerMethodField
 
+from wagtail.images.views.serve import generate_image_url
```
kaedroho, line 5: Is the view registered and using the [redirect](https://docs.wagtail.io/en/v1.9/advanced_topics/images/image_serve_view.html#making-the-view-redirect-instead-of-serve) mode? This prevents Wagtail from streaming images that are already rendered through the server.

jacobtoppm: This is correct, right?

https://github.com/newamericafoundation/newamerica-cms/blob/performance-improvements/newamericadotorg/urls.py#L45

home/templatetags/utilities.py
jacobtoppm: As far as I've read, Django doesn't guarantee an order, though it is likely to be pk? When I tried this out in silk, it looked like removing the order_by prevented Django from reevaluating the prefetched author queries (not here, but in the WeeklyArticleSerializer)

newamericadotorg/api/meta/views.py
```diff
@@ -7,6 +10,7 @@
 from newamericadotorg.settings.context_processors import content_types
 from newamericadotorg.api.program.serializers import ProgramSerializer, SubscriptionSegmentSerializer
 
+@method_decorator([cache_page(4*60*60)], name='get')
```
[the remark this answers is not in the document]
jacobtoppm: Good point, updated to use a couple of minutes now

newamericadotorg/api/author/serializers.py
kaedroho: Ahh, yep! Thanks!

home/templatetags/utilities.py
kaedroho: Oh, I didn't know this had an effect on prefetch! Nice one, thanks!

tomdyson:
@jacobtoppm do you have any before-and-after metrics, e.g. number of queries run, page response time?

jacobtoppm:
@tomdyson Approximately 650-700 queries down to 90ish, and (on my local environment, with silk installed, which can affect things) 7s-9s to 1s-2s.

tomdyson:
Thanks, @jacobtoppm, those sound very promising! Could we afford a longer cache for the meta API? 2000ms -> 100ms is a very dramatic improvement.

jacobtoppm:
@tomdyson Because some of what it's caching is updated via the CMS (programs for instance) and it's a relatively high traffic view, Karl suggested a shorter cache (30s) than what I'd added originally and we compromised at 2mins, but I'd definitely be up for extending it if we don't think the update frequency justifies the short cache duration. It's a combination of about pages, programs, content type pages, and subscriptions that get returned, and the latter I don't have a good feeling for how often are updated/how fast you'd want them to respond to an update. @nmorduch, do you have any thoughts on this?

kaedroho:
The current value of 2 minutes eliminates 96% of the traffic to that view (minimum, this would be even higher at peak times). We could increase it more, but the returns of increasing it diminish rapidly the higher we go. For example, if we now double it to 4 minutes, we would only gain an extra 2% hit rate.

jacobtoppm:
I suppose it depends whether we're going by hit rate or average request time - you're right about the hit rate improvement being slight though. Taking the cached view as somewhere between 100ms and 200ms, doubling the cache time gives us somewhere in the range of 13-20% reduction in average request time. Not huge, but noticeable.

nmorduch:
Is this good to merge?

jacobtoppm:
Yes, for sure. The cache length is easily adjustable if we decide it's needed later.

nmorduch:
I'm merging this to test on staging!

PR #1545, opened 2020-08-20
Remove option to copy pages

### LINKED ISSUE #1395: [2-3hrs] Remove or hide option to copy (duplicate) pages
@mariaelkin:
>Users will duplicate pages rather than making new child pages, which leads to errors. Rather than trying to fix it, I think we should hide the "copy" option from non-admin accounts or all accounts

chigby:
Fixes #1395

This change has two effects:

1. Removes the "Copy" option from the "More" dropdown menus on the page listing for everyone except superusers.

2. Installs a hook that rejects page copy actions from anyone except superusers. This fully prevents anyone in the disallowed group from copying a page, even by means other than now-removed Copy option.

### DIFF
--- a/home/wagtail_hooks.py
+++ b/home/wagtail_hooks.py
@@ -1,11 +1,19 @@
 from django.utils.safestring import mark_safe
 from django.utils.html import format_html, format_html_join
 from django.conf import settings
+from django.http import HttpResponseForbidden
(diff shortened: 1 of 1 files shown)


kaedroho (approved):
Looks good to me!

PR #1547, opened 2020-08-21
Update wagtail from 2.8 to 2.10

### LINKED ISSUE #1531: [2hrs] Update to Wagtail 2.10

chigby:
Fixes #1531 

## Summary of changes

In 2.9, `SiteMiddleware` was deprecated.  We weren't using `request.site` anywhere, so I removed the middleware itself as it's not needed.

In 2.10:

1. `get_document_model` import path changed
2. `<div class="rich-text">` wrappers removed from rich text.  Because we are using `.rich-text` selectors in our CSS, I've chosen to maintain the old behavior by adding a line to `INSTALLED_APPS`.  In the future we may want to change this, but I think that would require a more extensive review of rich text rendering across all the site content than I am capable of giving right now.

See this link for more information:

https://docs.wagtail.io/en/stable/reference/contrib/legacy_richtext.html#legacy-richtext

## Other considerations

In 2.10, the [workflow for moderation](https://docs.wagtail.io/en/stable/releases/2.10.html#move-to-new-configurable-moderation-system-workflow) seems to have changed substantially. It may be worth considering if these changes are applicable to users of our site, and what adaptations may have to be made.

Also, 2.10 drops support for python 3.5. We are using 3.6, which is also getting somewhat old. It may be worth considering upgrading to python 3.7 or 3.8 in the near future.

### DIFF
--- a/home/templatetags/top_menu.py
+++ b/home/templatetags/top_menu.py
@@ -1,6 +1,7 @@
 from datetime import date
 from django import template
 from django.conf import settings
+from wagtail.core.models import Site
 
 from programs.models import Program
 
@@ -10,7 +11,7 @@ register = template.Library()
 def get_site_root(context):
     # NB this returns a core.Page, not the implementation-specific model used
     # so object-comparison to self will return false as objects would differ
-    return context['request'].site.root_page
+    return Site.find_for_request(context['request']).root_page
 
 
 def has_menu_children(page):
--- a/newamericadotorg/settings/base.py
+++ b/newamericadotorg/settings/base.py
@@ -53,6 +53,7 @@ INSTALLED_APPS = [
     'wagtail.images',
     'wagtail.search',
     'wagtail.admin',
+    'wagtail.contrib.legacy.richtext',
     'wagtail.core',
     'wagtail.contrib.styleguide',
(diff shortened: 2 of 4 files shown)


nmorduch:
- Let's not worry about the moderation flow
- Let's check whether packages-not-used-by-Wagtail say they don't work with Python 3.8 (or 3.7) before updating Python. (Packages used by Wagtail should already be fine)

kaedroho:
We need to update this line https://github.com/newamericafoundation/newamerica-cms/blob/06be3c8b49576b2d8bdbc4d426fb9a74a995835e/home/templatetags/top_menu.py#L13 to no longer reference `request.site` (and instead use `Site.find_for_request`). See https://docs.wagtail.io/en/stable/releases/2.9.html#sitemiddleware-and-request-site-deprecated

kaedroho:
Just released: https://docs.wagtail.io/en/stable/releases/2.10.1.html

chigby:
Good catch on the `request.site` line! I've updated that code (though, it is possible that the `get_site_root` tag is not actually being used anywhere on the site right now) and bumped wagtail to use the latest, hot-off-the-presses release.

kaedroho (approved):

PR #1548, opened 2020-08-21
Add fabric task to copy heroku production data to staging

### LINKED ISSUE #1401: [2hrs] Syncing stage database with production
Hi @nmorduch just capturing this issue here, 

I'd suggest updating the stage database by downloading a backup from production and uploading it to staging as per the [heroku docs state](https://devcenter.heroku.com/articles/heroku-postgres-import-export) with the heroku cli, or is the request here to setup a sync task?

It's a good idea to configure a [fabric ](http://www.fabfile.org/) script to do this so it's easier going forward (we usually manage it this way). I think about 1 - 2 hours is enough time to configure that, this means we can sync staging with production on demand rather than a scheduled solution which may disrupt any current work in progress. 

Thanks

chigby:
This pull request adds a fabric task `sync-staging-from-production` which runs the [`heroku pg:copy`](https://devcenter.heroku.com/articles/heroku-cli-commands#heroku-pg-copy-source-target) command to copy the entire database from the production app into the staging app.

Fixes #1401 

### Notes

* The `fab sync-staging-from-production` command currently requires a user confirmation. I think this is all right if we mostly intend this command to be run by people, given that it is a destructive operation and it might be best to avoid accidentally overwriting data. However, if we want to automate this command, it might be best to alter the command to not require that confirmation step.

* At a technical level, the function `copy_heroku_database` uses required/enforced keyword arguments because I worry about the consequences of accidentally switching the `source` and `destination` arguments if they're supplied positionally. It's a slightly obscure feature though, and maybe the codebase doesn't need it. I suppose another option is to hard-code this copy operation. Either way, feedback on this is welcome and I'm fine to change it to something else.

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -74,6 +74,16 @@ def pull_staging_images(c):
     pull_images_from_s3_heroku(c, STAGING_APP_INSTANCE)
 
 
+@task
+def sync_staging_from_production(c):
+    """Copy database from production to staging"""
+    copy_heroku_database(
+        c,
+        destination=STAGING_APP_INSTANCE,
+        source=PRODUCTION_APP_INSTANCE,
+    )
+
+
 #######
 # Local
 #######
@@ -143,6 +153,17 @@ def open_heroku_shell(c, app_instance, shell_command="bash"):
     local(
         "heroku run --app {app} {command}".format(
             app=app_instance, command=shell_command
+        )
+    )
+
+
+# The single star (*) below indicates that all the arguments
+# afterwards must be given as keywords.  See PEP 3102 to learn more.
+def copy_heroku_database(c, *, source, destination):
+    check_if_logged_in_to_heroku(c)
+    local(
+        "heroku pg:copy {source_app}::DATABASE_URL DATABASE_URL --app {destination_app}".format(
+            source_app=source, destination_app=destination
         )
     )
 
--- a/fabfile.py
+++ b/fabfile.py
@@ -82,6 +82,11 @@ def sync_staging_from_production(c):
         destination=STAGING_APP_INSTANCE,
         source=PRODUCTION_APP_INSTANCE,
     )
+    sync_heroku_buckets(
+        c,
+        destination=STAGING_APP_INSTANCE,
+        source=PRODUCTION_APP_INSTANCE,
+    )
 
 
 #######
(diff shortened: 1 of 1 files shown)


fabfile.py
```diff
@@ -147,6 +157,15 @@ def open_heroku_shell(c, app_instance, shell_command="bash"):
     )
 
 
+def copy_heroku_database(c, *, source, destination):
```
nmorduch, line 160: Let's add a comment here noting what the `*` means, just in case

chigby:
Updated this PR with code to also sync media alongside the DB, but again, I have not actually tried it. Except for with the `--dryrun` flag on the media syncing command.

chigby:
@kaedroho Question for you: I've set up this command to sync media from production to staging, but I've seen that the other fabric commands distinguish between "media" and "images". Is this something we ought to be concerned with here as well, or is copying all media the way to go in this case?

kaedroho:
> Question for you: I've set up this command to sync media from production to staging, but I've seen that the other fabric commands distinguish between "media" and "images". Is this something we ought to be concerned with here as well, or is copying all media the way to go in this case?

I think copying all media is the best way to go.

For images, you can get away with copying just `original_images` as Wagtail will regenerate all the renditions in the `images` folder (if you delete all the rendition records in the database as well). Also, some people don't use documents, but I think we use them quite a bit on this site.

fabfile.py
```diff
@@ -147,6 +162,40 @@ def open_heroku_shell(c, app_instance, shell_command="bash"):
...
+        "heroku pg:copy {source_app}::DATABASE_URL DATABASE_URL --app {destination_app}".format(
+            source_app=source, destination_app=destination
+        )
+    )
+
+
+# The single star (*) below indicates that all the arguments
+# afterwards must be given as keywords.  See PEP 3102 to learn more.
+def sync_heroku_buckets(c, *, source, destination):
+    destination_bucket_name = get_heroku_variable(
```
kaedroho, line 179: Could we add a check to make sure the production bucket can't accidentally be wiped? just in case something was misconfigured somewhere.

eg:
```python
if destination_bucket_name == 'newamericadotorg':
    raise UserError("Production bucket used as destination for sync_heroku_buckets. Please delete this check if you're absolutely sure you want to do this")
```

kaedroho (approved):
Looks good to me! I added a comment about a check we could optionally add to catch any configuration mistakes.

nmorduch:
Merging but I'll open an issue to add that check.

fabfile.py
nmorduch: This is #1552

PR #1558, opened 2020-09-24
Improve production-to-staging sync

### LINKED ISSUE #1552: Add check to fabric task for copying production → staging, to make sure we never accidentally overwrite the production bucket
Could we add a check to make sure the production bucket can't accidentally be wiped? just in case something was misconfigured somewhere.

eg:
```python
if destination_bucket_name == 'newamericadotorg':
    raise UserError("Production bucket used as destination for sync_heroku_buckets. Please delete this check if you're absolutely sure you want to do this")
```

_Originally posted by @kaedroho in https://github.com/newamericafoundation/newamerica-cms/pull/1548#discussion_r483584166_

chigby:
Refs #1401 

This PR improves (or attempts to improve) the `sync-staging-from-production` fabric task by improving performance (the time it takes the command to run overall) and improved feedback to the user. Specifically:

1. Printing what media directories are being synced. These commands can take a long time so we should at least see a message stating so.
2. Restricted the number of files being synced by only syncing documents and original images. This should make it not take as long.
3. Use `aws s3 sync --size-only` to reduce the number of operations needed per file to be synced, which I think might improve performance (though I have not measured this). This option makes file comparisons based on size only, rather than a hash of the file's contents.

### Other notes

* Fixed #1552 
* I'm not sure S3 replica rules are what we want, it looks like these are for creating a system to automatically and immediately copy data from one S3 bucket to another, rather than doing so on-demand. I feel like having files automatically replicated to staging might cause problems?

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -3,7 +3,7 @@ from invoke.exceptions import Exit
 from invoke.tasks import task
 
 PRODUCTION_APP_INSTANCE = "newamericadotorg"
-STAGING_APP_INSTANCE = "na-staging"
+STAGING_APP_INSTANCE = "na-develop"
 
 LOCAL_MEDIA_FOLDER = "/vagrant/media"
 LOCAL_IMAGES_FOLDER = "/vagrant/media/original_images"
@@ -173,7 +173,7 @@ def open_heroku_shell(c, app_instance, shell_command="bash"):
 def copy_heroku_database(c, *, source, destination):
     check_if_logged_in_to_heroku(c)
     local(
-        "heroku pg:copy {source_app}::DATABASE_URL DATABASE_URL --app {destination_app}".format(
+        "heroku pg:copy {source_app}::DATABASE_URL DATABASE_URL --app {destination_app} --confirm {destination_app}".format(
             source_app=source, destination_app=destination
         )
     )
@@ -216,7 +216,6 @@ def sync_heroku_buckets(c, *, source, destination, folders=[]):
             folder='',
         )
         aws(c, aws_cmd, destination_access_key_id, destination_secret_access_key)
-
 
 ####
 # S3
--- a/fabfile.py
+++ b/fabfile.py
@@ -196,13 +196,12 @@ def sync_heroku_buckets(c, *, source, destination, folders=[]):
         c, source, "S3_BUCKET_NAME"
     )
 
-
     # The `--size-only` flag means we don't have to compute the md5
     # hash of every media file, potentially increasing performance
     cmd_template = "s3 sync --size-only --delete s3://{source_bucket_name}/{folder} s3://{destination_bucket_name}/{folder}"
     if folders:
         for folder in folders:
-            print('Syncing media folder `{}` (this may take a while).'.format(folder))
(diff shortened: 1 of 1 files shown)


fabfile.py
```diff
@@ -86,7 +86,13 @@ def sync_staging_from_production(c):
         c,
         destination=STAGING_APP_INSTANCE,
         source=PRODUCTION_APP_INSTANCE,
+        folders=['original_images', 'documents']
     )
+    # The above command just syncs the original images, so we need to
+    # delete the contents of the wagtailimages_renditions table so
+    # that the renditions will be re-created when requested in the
+    # staging environment.
+    delete_staging_renditions(c)
```
jacobtoppm, line 95: As this is a separate step, maybe we could stick another print in there, just to give some more feedback as to what's going on?

jacobtoppm (changes requested):
Generally looks great! Just a couple of minor suggestions

fabfile.py
```diff
@@ -188,12 +196,27 @@ def sync_heroku_buckets(c, *, source, destination):
...
-    )
-
-    aws(c, aws_cmd, destination_access_key_id, destination_secret_access_key)
 
+    # The `--size-only` flag means we don't have to compute the md5
+    # hash of every media file, potentially increasing performance
+    cmd_template = "s3 sync --size-only --delete s3://{source_bucket_name}/{folder} s3://{destination_bucket_name}/{folder}"
+    if folders:
+        for folder in folders:
+            print('Syncing media folder `{}` (this may take a while).'.format(folder))
```
nmorduch, line 205: Can we add to "this may take a while" a "you'll need to keep your computer awake and connected" (assuming that you do need to)

nmorduch (changes requested):
- [ ] The original images are synced, but there aren't any new renditions in the bucket
  - [ ] I also think it might make sense to delete existing renditions from the bucket, but that seems like not really a big deal
- [ ] Let's add a `sync_develop_from_production` (fine with me if it's a copy-pasted duplicate tbh, for time efficiency)

fabfile.py
```diff
@@ -321,5 +344,14 @@ def delete_local_renditions(local_database_name=LOCAL_DATABASE_NAME):
         pass
 
 
+def delete_staging_renditions(c):
+    check_if_logged_in_to_heroku(c)
+    local(
+        'heroku pg:psql --app {app} -c "DELETE FROM wagtailimages_rendition;"'.format(
```
nmorduch, line 350: @chigby just figured out that this should be deleting `home_customrendition` and should be for `delete_local_renditions` as well

chigby:
All right, I _think_ I've taken care of all the great suggestions offered here, plus the problem with the image renditions deleting from the wrong table. Should be ready to re-review.

jacobtoppm (commented):
Unfortunately I don't have access to staging, so had to test this by changing the staging app name to na-develop. However, I ran into two issues doing this:
- It asked for confirmation by entering the app name, but submitted the app name after entering just one character. To get around this, I needed to copy and paste the app name.
- After "Copying... done" I got no further feedback and no continuation despite leaving for about 45 minutes. It doesn't seem to move onto the bucket syncing task. Potentially this is related to my permissions or the fact I'm doing this with "develop" rather than staging - have you encountered this before?

nmorduch:
I'm having both of the same problems as @jacobtoppm

Issue #1539, opened 2020-08-19
Create a "Figure" streamfield block


nmorduch:

nmorduch:
@chigby to make a new the model, `FigureBlock` should have:

- title (plain text)
- text (rich text. help text: "Displays below the title") (can we limit the length?)
- caption (rich text. help text: "Displays like a footer") (can we limit the length?)
- width (same options as on DatawrapperBlock)
- streamfield: Can have 0-8 (inclusive) of: CustomImageBlock, DatawrapperBlock, IframeBlock, DatavizBlock
- columns (number: 1-4 inclusive)

And it should be in the group 'Embeds'

chigby:
I've made a `FigureBlock` with these fields.. I'm not what models to enable it for? Or is that something you were planning on doing?

chigby:
@nmorduch Check out the draft PR here for the block code and a suuper basic template (and let me know if there are issues or enhancements you need to flesh out the html/css situation)

https://github.com/newamericafoundation/newamerica-cms/pull/1559

PR #1559, opened 2020-09-25
FigureBlock support in streamfields

chigby:
Refs #1539

WIP


nmorduch:
Closing this, and it'll still be here if we want to come back to it later.

Issue #1530, opened 2020-05-07
[Search] Show images and content types for all content types


nmorduch:
Currently, images and page types don't show for programs, subprograms, topics, about pages, etc.

The content type is important because it's confusing when you have two pages that look the same but one is a Subprogram page and one is the About page for that subprogram.

The image is a little less important but it's p ugly without.

Here's a screenshot from the dev server (with breadcrumbs)

<img width="1265" alt="image" src="https://user-images.githubusercontent.com/1051450/81324851-8a068080-9065-11ea-9c4d-46c04e4abddf.png">

nmorduch:
Accidentally closed even though still images still need to be added to some page types

PR #1561, opened 2020-09-29
Search content type

jacobtoppm:
Addresses https://github.com/newamericafoundation/newamerica-cms/issues/1530

- Display content type even if no program is found for publication listing cards, including search
- Find the program even for non Posts and for models linked to Program via tree structure rather than `parent_programs`
- Look for story image via `getattr` to find story images on non Posts

However, some page types still don't have a story image, and I wasn't sure if this was something you wanted to add as part of this, so not all pages will show images in search.

### DIFF
--- a/home/models.py
+++ b/home/models.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 from django.db import models
 from django.apps import apps
 from django.shortcuts import redirect
+from django.utils.functional import cached_property
 from datetime import datetime
 from pytz import timezone
 from django.db.models import Q
@@ -368,9 +369,13 @@ class ProgramSimplePage(AbstractSimplePage):
     parent_page_types = ['programs.Program', 'programs.Subprogram']
     subpage_types = ['home.RedirectPage']
 
+    @cached_property
+    def program(self):
+        return self.get_parent().specific
+
     def get_context(self, request):
         context = super(ProgramSimplePage, self).get_context(request)
-        context['program'] = self.get_parent().specific
+        context['program'] = self.program
 
         return context
 
@@ -388,7 +393,7 @@ class ProgramAboutHomePage(ProgramSimplePage):
 
     def get_context(self, request):
         context = super().get_context(request)
-        context['program'] = self.get_parent().specific
+        context['program'] = self.program
 
         if getattr(request, 'is_preview', False):
             program_context = context['program'].get_context(request)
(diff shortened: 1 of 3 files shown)


home/models.py
```diff
@@ -388,7 +393,7 @@ class Meta:
 
     def get_context(self, request):
         context = super().get_context(request)
-        context['program'] = self.get_parent().specific
+        context['program'] = self.program
```
kaedroho, line 396: Looks like `context['program']` is already being set in `super().get_context(request)`

home/models.py
```diff
@@ -406,9 +411,13 @@ class ProgramAboutPage(ProgramSimplePage):
...
         verbose_name = 'Program About Page'
 
+    @cached_property
+    def program(self):
+        return self.get_parent().get_parent().specific
+
     def get_context(self, request):
         context = super(ProgramSimplePage, self).get_context(request)
-        context['program'] = self.get_parent().get_parent().specific
+        context['program'] = self.program
```
kaedroho, line 420: As above, but you might need to remove the params in `super()` first to make it work

home/models.py
```diff
@@ -406,9 +411,13 @@ class ProgramAboutPage(ProgramSimplePage):
     class Meta:
         verbose_name = 'Program About Page'
 
+    @cached_property
+    def program(self):
+        return self.get_parent().get_parent().specific
```
kaedroho, line 416: Would it be possible to replace this with something like: ``Program.objects.ancestor_of(self).get()``?

newamericadotorg/api/search/serializers.py
```diff
@@ -23,13 +23,18 @@ def to_representation(self, obj):
...
             data['position'] = obj.position_at_new_america
         elif isinstance(obj, Post) or type(obj) == Event:
-            data['story_image'] = generate_image_url(obj.story_image, 'max-300x240') if obj.story_image else None
             data['date'] = obj.date
-            data['programs'] = PostProgramSerializer(obj.parent_programs, many=True).data
 
         if isinstance(obj, Post):
             data['authors'] = AuthorSerializer(obj.post_author, many=True, context=self.context).data
+        
+        data['story_image'] = generate_image_url(obj.story_image, 'max-300x240') if getattr(obj, 'story_image', None) else None
```
kaedroho, line 31: I think it would be better if we could create some subclasses of this serializer for more specific types, but this is fine for now!

kaedroho (changes requested):

home/models.py
jacobtoppm: Good point, thanks!

home/models.py
jacobtoppm: Yes, but it needs to include subprograms, so I've used `AbstractProgram` instead - changed in latest commit.

newamericadotorg/api/search/serializers.py
jacobtoppm: I definitely agree! It was more a question of scope for now, since I'd need to familiarise myself more with all the different types in use and which would require a more specific serialiser, but it's definitely the way to go

kaedroho (approved):

nmorduch:
Merged in f8afaa229e8168c636adc386410a012f99d19f0d (see #1563)

Issue #1522, opened 2020-04-27
[𝛆] Stop using absolute url for timeline js


nmorduch:
Examples of instances where we're using absolute urls and shouldn't be:

- [x] https://github.com/newamericafoundation/newamerica-cms/blob/master/report/templates/report/pdf.html#L31
- [x] https://github.com/newamericafoundation/newamerica-cms/blob/master/report/templates/report/pdf.html#L44
- [ ] https://github.com/newamericafoundation/newamerica-cms/blob/master/newamericadotorg/templates/blocks/timeline.html#L67
- [x] We should do a project search to find any others, as well, and make them settings or similar.

nmorduch:
check out EXTRA_STATIC_URL in settings, can pull that into the template

nmorduch:
This has been causing problems while we're testing the Wagtail 10.2 update

chigby:
Note: this was changed _to_ the absolute URL in #1518

PR #1568, opened 2020-10-20
Pdf images

jacobtoppm:
Remove hardcoded urls from pdf templates, which were preventing image urls from being generated properly for staging images.


nmorduch:
Will close #1522 if we check for other instances across the site

nmorduch (approved):
Doesn't get the timeline js but i think that's much less urgent. Works locally.

PR #1572, opened 2020-10-21
1538 pdf image height

### LINKED ISSUE #1538: [1hr] Large images getting cut off in report PDF

nmorduch:
Fixes #1538 

Notes:
- I pulled in Jacob's fix for PDF images so that I would be able to view images in order to test them.
- PDF generation wasn't working for me locally but I tested by printing to PDF in my browser (from the print view, which is when you add `/print/` to the end of a report url). 
- For the `max-height` I used the largest value that l thought should work according to the page margins indicated by the CSS, but I'm not entirely sure it's right because I wasn't able to test properly.
- Arguably I maybe should have put the max height on `img` not `.figure__image > img` but this will make a big difference regardless. (Maybe we would need it on `img` if people insert images inside a rich text block? But I don't think we use that much and I'm frustrated with my inability to test this properly, at this point.)

To test:
- Starting on `main`, upload an image that is tall and not awfully wide. My test image was 1387 × 3147px.
- Verify that your image extends off the bottom of the page when you generate the pdf
- Switch to `1538-pdf-image-height` and regenerate the pdf
- Verify that your image no longer extends off the bottom of the page

### DIFF
--- a/report/templates/report/figure.html
+++ b/report/templates/report/figure.html
@@ -12,7 +12,7 @@
     </div>
   {% endif %}
   <div class="figure__image">
-      <img src="https://s3.amazonaws.com/newamericadotorg/{{value.image.file.name}}"/>
+      <img src="{{ value.image.file.url }}"/>
   </div>
   {% if value.image.caption or value.image.source %}
     <figcaption>
--- a/report/templates/report/figure_iframe.html
+++ b/report/templates/report/figure_iframe.html
@@ -3,7 +3,7 @@
 <figure class="figure {{value.fallback_image_align}} {{value.fallback_image_width|temp_image_width_map}}">
 
   <div class="figure__image">
-      <img src="https://s3.amazonaws.com/newamericadotorg/{{value.fallback_image.file.name}}"/>
+      <img src="{{ value.fallback_image.file.url }}"/>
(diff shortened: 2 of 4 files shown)


chigby:
So, I did manage to get this working locally, though only after a bit of confusion where my image wasn't showing up in the PDF because it was a JPEG. I fixed this by installing `libgdk-pixbuf2.0-0` on vagrant (it's apparently what weasyprint uses for this). 

@nmorduch The way I tested the PDF generation was to add `/pdf/render/` to the report URL. Does that work for you?

nmorduch:
Oh, good call on testing that way, and that did indeed work for me (both the pdf rendering worked and the change worked). What wasn't working was getting a new PDF when I saved the report.

nmorduch:
Not sure why it says CI hasn't passed yet, considering that it definitely has:
https://travis-ci.org/github/newamericafoundation/newamerica-cms/builds/737824714
https://travis-ci.org/github/newamericafoundation/newamerica-cms/builds/737827844

chigby (approved):
I tried this out locally with an extra-tall image and the result was not being cut off by a page-break in the final PDF. So by that measure, I think this works.

PR #1573, opened 2020-10-22
Add GDK-Pixbuf to the vagrant provision script

chigby:
This pull request adds the GDK-Pixbuf library to the default vagrant provisioning script. This package is needed for weasyprint to support embedded images in the JPEG format.


jacobtoppm (dismissed):
Looks good to me!

Issue #1562, opened 2020-09-29
[Search] Separate results into groups by page type


nmorduch:
Separate results by page type, in a few groups of page types. I'll be referring to these each of these as a "group" in this issue,

- **Programs, Initiatives, and Projects** (programs and subprograms, list programs first)
- **Publications** (see [Publications page](https://www.newamerica.org/publications/))
- **Events** (see [Events page](https://www.newamerica.org/events/))
- **People** (search only names, see [Our People page](https://www.newamerica.org/our-people/), which has current people)
- **Other pages** (everything else, searched however the search already works)

Details:
- [ ] I think we should base this on #1494 
- [ ] Not sure whether we should search each group separately from the others or do the search all at one time.
- [ ] For each group, have the title (the text bolded in the list above) then display the results for that type. Here's a screenshot of adding in an `h2` as the first child of `.program__publications-list-wrapper` (which I would have one of that list and its wrapper for each group)
  <img width="1128" alt="image" src="https://user-images.githubusercontent.com/1051450/94580471-f9b11e00-0247-11eb-94d5-58b86611a30d.png">
- [ ] Only display the groups that have results. (And then list at the bottom of the page any buckets that didn't have results.)

nmorduch:
For overall ordering, let's switch to making no date sort as though its date were 2000-01-01, which is earlier than everything except 1 book.

Let's have the groups, in the following order, paginated in groups of 6, except only 4 upcoming events. For pagination, either the "next" button should flip the page for every category at the same time (possibly easier? not ideal) or there should separate pagination for each category (probably fine)

- **Programs, Initiatives, and Projects** 
  - Programs should come first, then Subprograms, and it might be useful to have a class on the programs or something so that there can be a visual distinction
- **Upcoming Events**
  - I just realized that this was something we should put ahead of publications, and then putting past events in with publications
  - show only 4 of these
  - hide if none
- **Publications and Past Events**
  - Put In Depth sections last (but keep In Depth main pages)
- **People**
- **Other pages** 
  - hide if none
  - Exclude from each results entirely (there are probably more but here's what i've noticed so far)
    - mailing list segment
    - redirect pages


And here's a fairly hacky mockup. I stand by none of the choices about how the results display, but i figured i'd use screenshots in a relatively attractive way. You can just look at it for inspiration leave and display of the results in the same format they are now, and without adding anything about columns

[...]

PR #1575, opened 2020-10-28
Search result groups

chigby:
Groups search results into groups by content type on the search result page

Refs #1562

### DIFF
--- a/newamericadotorg/assets/react/home-panels/pages/Search.js
+++ b/newamericadotorg/assets/react/home-panels/pages/Search.js
@@ -7,7 +7,7 @@ import { PublicationListItem, Person } from '../../components/ContentCards';
 import { LoadingDots } from '../../components/Icons';
 
 
-function SearchResultList({results, header, isFetching, hasNext, hasPrevious, fetchNext, fetchPrevious}) {
+function SearchResultList({results, header, bucket, isFetching, hasNext, hasPrevious, fetchNext, fetchPrevious}) {
   // Remember if we've tried to fetch before
   // This helps us differentiate between the initial state and "No results"
   const [hasFetched, setHasFetched] = useState(false);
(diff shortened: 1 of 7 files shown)


nmorduch (changes requested):
I'm running into a problem where, because this categorizes whatever comes in on a given page of search results, the first page has a lot of programs and no people, and the second page has a lot of people. I probably described unclearly what I was looking for, which was that it should have these 5 sections, and then under each section we should return the first x results if that section has any results at all (which was why I was imagining doing it on the backend, since that's where the pagination is). I'll update the issue with a mockup that gives a sense of how many of each type I'm looking to see (since i think it will be different for different types) along with any ordering we want within the category (which, to be fair, is just now occurring to me as being desirable).

newamericadotorg/assets/react/home-panels/pages/Search.js
```diff
@@ -6,6 +6,37 @@ import * as QueryString from 'query-string';
...
+      collection.push(item);
+    }
+  });
+  return map;
+}
+
+function classifyResult(result) {
+  let type = result.content_type.api_name;
+
+  if (type == "article" || type == "blogpost" || type == "book" || type == "event" || type == "issueortopic" || type == "podcast" || type == "policypaper" || type == "pressrelase" || type == "quoted" || type == "weeklyarticle" || type == "report" || type == "indepth") {
```
nmorduch, line 27: Looks like events are going into this category so not making it to the "events" category, and "issueortopic" here should go into "other pages"

chigby:
@nmorduch Updated this PR to be more in line with what you're describing. The search page now separately searches each of the buckets and returns separately-paginated result buckets.

chigby:
Further updated the PR to reduce the number of results per page on each bucket to 6. Also made it customizable per-bucket.

nmorduch:
@chigby can you review my commit? and this should be deployed to the dev server soon

nmorduch (approved):

PR #1580, opened 2020-11-17
Clean up setup instructions in README

nmorduch:


nmorduch:
We're planning to stop using vagrant anyway with #1642

PR #1593, opened 2021-01-21
Add OtherPost to possible pages usable as featured publications

### LINKED ISSUE #1586: [𝛆] Allow featuring "Other Post" pages on Topic pages
Similar to #1542, allow featuring "Other Post" on the topic page
Probably just need to add it here: https://github.com/newamericafoundation/newamerica-cms/blob/66425e8e2cff9c0930ff6a7f82166a7f8ce9f90d/issue/models.py#L71-L83

Might also be good to check if there are other places "Other Post" isn't fully integrated, but I also wouldn't let the perfect be the enemy of the good here.

chigby:
Fixes #1586 

This change expands the list of allowed pages in a topic page's featured publications to include `OtherPost` pages.

### DIFF
--- a/issue/models.py
+++ b/issue/models.py
@@ -72,13 +72,13 @@ class IssueOrTopic(ProgramSimplePage):
         MultiFieldPanel([
             PageChooserPanel(
                 'featured_publication_1',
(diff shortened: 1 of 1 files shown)


nmorduch (approved):
Works locally!

PR #1615, opened 2021-02-16
Fix images in Resource Kit display

chigby:
See commit message for explanation of changes.


nmorduch (approved):
Works locally and looks great!

PR #1623, opened 2021-02-25
Elasticsearch 5 upgrade

### LINKED ISSUE #1613: [3h] Upgrade Elasticsearch to v5
Required for Wagtail 2.12

chigby:
Fixes #1613 

Will require some kind of re-indexing and/or migration of the Elastic Search service on Heroku.

I'll also note that we may want to consider also upgrading to ES 6 at some point, as ES 5 is quite out of date as well.


chigby:
Looked this over again and I still think we're in good shape here. I think the next step is to test it out on dev or staging with the upgraded ES heroku add-on.

nmorduch:
To do: add description how how to upgrade on the server

PR #1625, opened 2021-02-25
Fix polling dashboard

chigby:
Use correct attribute association for title of associated commentary.


nmorduch (approved):
PEP8 compliance much appreciated, as is putting it in its own commit :)

I'm not able to replicate the bug locally atm, but I also get no bugs on this branch and I can't imagine it will make any new bugs at least, so I'm approving it.

PR #1631, opened 2021-02-26
Rework search results page to support sortable buckets

### LINKED ISSUE #1608: [1h] Move empty search buckets to bottom of page

chigby:
Fixes #1608 

This required slightly more surgery than I thought it would, because in order to sort search buckets, we must consider them all together, meaning we must "move the state up" from the bucket to the search container.  What's happening here is somewhat straightforward, but it's a lot of code churn.  Essentially the child components of `Search` have had their behavior moved into the top-level component so we can see how many results they have all at once.  So we store the relevant data in an array and it flows into the children, whose job now is just to render it.

I hope I didn't miss anything here, but it worked well for me, locally.


nmorduch:
Closed in favor of #1632

PR #1632, opened 2021-03-12
Use flexbox ordering to sort search buckets

### LINKED ISSUE #1608: [1h] Move empty search buckets to bottom of page

nmorduch:
- Moves empty buckets to end of list
- Use fragment for search bucket so that search results won't be wrapped in a div

Closes #1608
Alternative to #1631


chigby (approved):
Solution works great for me locally, and much less code modification than the other PR. Let's go with this solution. Thanks for getting it in there.

PR #1633, opened 2021-03-16
Use flexbox in order to avoid safari bug

nmorduch:
Bug: full-bleed header image for reports not showing at all on safari, because of a bug that makes height:100% not work with flexbox (see https://stackoverflow.com/a/33644245)

Full-bleed, single page: https://www.newamerica.org/political-reform/reports/choosing-to-vote-as-usual/
Full-bleed, multi-page: https://www.newamerica.org/education-policy/reports/transforming-financing/


chigby (approved):
Works to fix the problem, and the result is also cleaner code.

PR #1640, opened 2021-04-20
Hide redirect projects in search

### LINKED ISSUE #1639: Search: subprogram pages that redirect should not show up in search
see https://www.newamerica.org/search/?query=Cyber+Citizenship

There are two subprograms that show up. they both _look_ the same, but they link to different places. One is in ISP and isn't a real page, just redirects to the EPP one, and it should not show up in the list. (the "redirect" subprograms are imo a bad setup, but i don't think it's likely to be worth changing that setup to fix this)

chigby:
Fixes #1639 

In order for this to change take effect on staging/production/etc., I believe we have to run the `update_index` management command.

### DIFF
--- a/programs/models.py
+++ b/programs/models.py
@@ -383,6 +383,11 @@ class Project(Subprogram):
         ObjectList(Subprogram.settings_panels, heading='Settings', classname='settings'),
     ])
 
+    @classmethod
+    def get_indexed_objects(cls):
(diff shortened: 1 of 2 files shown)


nmorduch (approved):
Finally finished testing this and it's working great! And thanks for making the reordering its own commit!

Issue #1637, opened 2021-04-16
New publication type: Brief


nmorduch:
Blocked by #1636

New publication type called Brief
- publication needs to be approved (#1636)
- [x] limit to 3k words ("brief" lol)
- [x] can have attached pdf
- [x] most similar to either a blog post or a policy paper
- [x] full-width DW charts need to work, etc (#1571)
- ~~pare down blocks available so there no legacy blocks (new page type doesn't need legacy things)~~

@chigby will do BE then @nmorduch will do FE

chigby:
A couple questions:

1. When we limit it to 3,000 words, does that mean specifically the word count of the `body` field? As in, we want to count the words across all the blocks in the body and if the total is over 3k, the page is invalid and won't save? Or something else?
2. Do you have a list of which blocks are considered legacy blocks?

PR #1647, opened 2021-04-22
Brief pages

chigby:
Refs #1637


brief/models.py
```diff
@@ -0,0 +1,78 @@
...
+        null=True,
+        blank=True,
+    )
+
+    content_panels = Post.content_panels + [
+        StreamFieldPanel('attachment'),
+    ]
+
+    def word_count(self):
+        countable_block_types = {'paragraph', 'introduction', 'heading'}
```
nmorduch, line 28: We should also count: 
- Resource kit description
- Table content I guess?
- Everything we're already counting, but as children of a Panel

nmorduch (changes requested):
My changes to the models files are nitpicky, just was copying from a different app's models.

To do:
- rebase off `main` (has a conflict in the imports so i'm leaving it for you. and then tests will pass)
- update things we're counting for word limit (see line comment)
- maybe make it have just one migration instead of both of these, but i don't know how much i care

nmorduch:
Actually, how about merging #1650 into this branch, and then add the brief publishing permissions to this branch after that?

brief/models.py
```diff
@@ -0,0 +1,85 @@
...
+    countable_block_types = {'paragraph', 'introduction', 'heading'}
+    count = 0
+    for block in body_block:
+        if block.block_type in countable_block_types:
+            text = getattr(block.value, 'source', block.value)
+            count += len(str(text).split())
+        elif block.block_type == 'resource_kit':
+            count += len(block.value.get('title', '').split())
+            count += len(block.value.get('description', '').split())
+            for block_type, data in block.value['resources'].stream_data:
```
nmorduch, line 22: I'm getting:
> **Exception Type:** ValueError
> **Exception Value:** too many values to unpack (expected 2)
> **Exception Location:** /vagrant/brief/models.py in word_count_body_block, line 22

Also, I don't really need to count all the things in the resources. I would just keep lines 20 and 21 I think

nmorduch (changes requested):
See line comment

brief/wagtail_hooks.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+    if isinstance(context.get('page'), Brief):
+        settings = PublicationPermissions.for_request(request)
+        if request.user.is_superuser:
+            # superusers can always publish
+            return
+
+        if not request.user.groups.all() & settings.groups_with_brief_permission.all():
+            menu_items[:] = [
+                item for item in menu_items if item.name != 'action-publish'
```
nmorduch, line 23: I didn't catch this when testing reports, but my limited-permissions user can publish a new report or brief, just can publish edits to an existing one.

chigby: It's possible you need to restart your wagtail server locally. These hooks do not take effect except at the time wagtail starts up.

nmorduch: Didn't fix it :/ Can still make new ones. (Can't edit existing ones! Typo above was unclear)

nmorduch:
Word count situation is fixed!

brief/wagtail_hooks.py
chigby: I think I misread your comment earlier. Just pushed a fix for limiting the permissions when creating a new report or brief.

nmorduch: Prob misread it because i'd had miswritten it 

Also this is fixed now!

nmorduch (approved):
This is great! Going to put this on the dev server so other people can see how great it is :)

nmorduch (changes requested):
Wait a minute you lost my commits when you rebased this :p

nmorduch:
These two, of which the first is actually important:
https://github.com/newamericafoundation/newamerica-cms/commit/5c3e00c46d3c3e066967b706b15f0ad42f8e06c5
https://github.com/newamericafoundation/newamerica-cms/commit/0af73fe85be25880cb19a3a22041b48c62376a52

nmorduch:
for my review dismissal, should say: "added changes"

nmorduch (dismissed):
Cameron says he approves of my changes. This is currently in testing on the dev server.

nmorduch:
# Notes re testing on the dev server

## What I’ve set up in the admin

- I set it so “Moderators” can publish Briefs and Reports. This is in the settings menu, on the new “Publication permissions” page. We could also set those to no groups, and then it would only be “Administrators” who could publish reports & briefs.
- I made a test user who has “Editor” but not “Moderator” permissions. That user should not be able to publish any reports or briefs.
- I also made a workflow for moderator approval of reports and briefs in Asset Building, and reports in BLL just to show that it could be one workflow covering multiple programs.

### What will be different in production

- The workflow will need to apply to every program’s reports, and we’ll need to add their briefs homepages as they start having briefs homepages. (I don’t see a reason to a briefs homepage before a given program starts publishing briefs.) Fwiw, if we don’t add their reports/briefs to the workflow, they can still save drafts, they just don’t have a built-in way to submit them for moderation.
- We’ll need to make new permission groups for people who currently have admin permissions and need some permissions that they can’t get with the basic program permissions, but don’t need all the permissions. We should prob warn them when or before we do this.

## Testing

- https://na-develop.herokuapp.com/briefs/ you can filter for briefs, and the one there has two attachments.
- You can also see that brief here https://na-develop.herokuapp.com/asset-building/publications/
- Briefs are capped at 3000 words. We check some types of text and not others, but an approximate cap should be fine for our purposes. Try publishing something that’s too long, and see that the error says how many counted words you have.
- As the test user, you should not be able to publish new reports or briefs, nor edit existing ones
- Try out the moderation workflow. I would do this by having a private/incognito window where I’m logged in as the test user and submit a report or brief for moderation, then approve/reject from my regular account, and switch back to see what feels like for a user to get a page approved or rejected.

nmorduch:
Let's change this to 5k words also

nmorduch (approved):
@chigby approves latest commit

PR #1648, opened 2021-04-22
1634 the thread

### LINKED ISSUE #1634: [3hrs] Make a duplicate of The Weekly called The Thread
Make it so there can be a second publication using the Weekly models, which will be called The Thread. They will exist side by side, but we will not highlight The Weekly anywhere.

- [ ] There are some places where Weekly or The Weekly is hard coded and that's not going to work now.
- [ ] The navigation and the link the on homepage need to be for The Thread (probably not worth changing this from being hard-coded)
- [ ] On the Publications page, change the Weekly filter to be called The Thread, and exclude pages from the old Weekly

The new slug should be `the-thread` (which we'll need for all the hard-coding lol)

nmorduch:
Went over this with Maria to make sure all the text was what she wanted.

I would def go through this by commit given the first commit is duplicated and then a lot of it was find-and-replace (but please do glance through the find-and-replace because I did some of it manually).

I'm putting this on the dev server and will make the appropriate pages

Closes #1634


home/tests.py
```diff
@@ -176,12 +177,14 @@ def test_homepage_subpages(self):
...
             SubscribePage,
             Weekly,
+            Thread,
             RedirectPage,
             SubscribeHome,
             AllReportsHomePage,
             PublicationsPage,
             AllOtherPostsHomePage,
             AllWeeklyArticlesHomePage
+            AllThreadArticlesHomePage
```
chigby, line 187: Syntax: Need comma on previous line

nmorduch:
I rebased since deploying the the dev server, so it no longer has the homepage changes. Those are on #1649 so that we can wait to merge them

PR #1649, opened 2021-04-23
1634 the thread links

nmorduch:
Merge and deploy this when The Thread is ready to go live.

Changes the homepage to link to the thread. Also changes links in `top_menu.html` but they don't appear to show up anywhere.


chigby (dismissed):
Everything here seems all set to me.

chigby (approved):
Nav links look good to me! 👍

PR #1650, opened 2021-04-23
Report publish permissions

### LINKED ISSUE #1636: Set publishing permissions by page type
Ideally with the option to use workflows with this.

- [ ] Restrict who can publish Reports
- [ ] Require approval to publish Briefs

chigby:
Fixes #1636 

This PR adds a section to the site settings called "Publication Permissions" that allows admins to specify what groups have access to the "Publish" action for Report page and for Brief pages (which are forthcoming in #1647). If a user is not a superuser or in one of these groups, the publish action is removed from the action menu on report pages.


nmorduch (commented):
Looks good and works for me! We should not merge until we're ready to make this change live. And we'll want to put it on the dev server and test the new setup first.

nmorduch:
(I would approve but I want to wait in order to remind us that it should first be tested by people who will use it.)

PR #1652, opened 2021-04-27
Sync avatar images when syncing all media

nmorduch:
This is the silliest of pet peeve changes but it was bothering me because when there's an image set but no image to put there, it's just blank, whereas people who don't have an image set have a placeholder. so syncing the avatar_images folder (which has all of three image in it) fixes that. plus it's DRYer fwiw


chigby:
Investigate if this will let us preview the reports page

nmorduch:
lol that comment is in the wrong location @chigby hahaha

nmorduch:
@chigby says he approves

chigby (approved):
Looks great

PR #1653, opened 2021-04-30
Headless preview

### LINKED ISSUE #1635: [2hr] Headless preview
Might want to use https://github.com/torchbox/wagtail-headless-preview

Highest priority pages for this: reports, followed by program about pages ([eg](https://www.newamerica.org/fellows/about/))

chigby:
Fixes #1635 

So far:

1. Only applies to reports
2. Uses a hard-coded "localhost" URL. This is what the docs say to do, but it means it probably won't work outside of a local dev environment. There's probably a way around this, like setting up specific URLs for each of our environments, but it is not done yet.

### DIFF
--- a/newamericadotorg/assets/react/report/index.js
+++ b/newamericadotorg/assets/react/report/index.js
@@ -319,6 +319,10 @@ class Routes extends Component {
             render={() => <Redirect to={results.url} />}
           />
           <Route
+            path={`/h_preview`}
+            render={this.reportRender}
+          />
+          <Route
             path="/:program/reports/:reportTitle/:sectionSlug?"
             render={this.reportRender}
           />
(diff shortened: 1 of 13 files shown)


nmorduch:
>     2. Uses a hard-coded "localhost" URL. This is what the docs say to do, but it means it probably won't work outside of a local dev environment. There's probably a way around this, like setting up specific URLs for each of our environments, but it is not done yet.
We should be able to set a setting on Heroku for each environment, yeah?

nmorduch (changes requested):
I'm getting errors related to using an iframe :/
@chigby noted this is mentioned in https://github.com/torchbox/wagtail-headless-preview/issues/15

newamericadotorg/assets/react/report/index.js
```diff
@@ -318,6 +318,10 @@ class Routes extends Component {
             path={`/admin/pages`}
             render={() => <Redirect to={results.url} />}
           />
+          <Route
```
nmorduch, line 321: Should we delete the route above this? I think it might be to make things less look broken when preview doesn't work.

nmorduch (changes requested):
This is working for the landing page, but does not then work for the individual pages of the reports. We prob have to change the links in the nav not to be full urls, at least for the preview, and make it so that the routes then work.

nmorduch (approved):
This looks great! You'll need to resolve the conflicts, and we should make a (lower priority) ticket for about page previews since this only fixes the preview for report pages

chigby:
Agree re: about pages. The main reason I didn't roll that into this ticket is I wasn't sure exactly if this technique would work or not.

nmorduch (approved):

Issue #1396, opened 2019-08-19
Checkbox on subscribe page should default to checked


nmorduch:
https://www.newamerica.org/subscribe/

kevinhowbrook:
Look like we just need to know what checkboxes need to be the default for this, setting them as default can be configured in the UI by editing the subscribe page:

![subsribe checkbox example](https://i.imgur.com/N6gy9ed.png)

nmorduch:
Thanks, @kevinhowbrook :)

nmorduch:
@kevinhowbrook two things:

- i checked the boxes on the main subscribe page and it didn't actually make a difference
- we also want to do this for the program subscribe pages, where the option isn't available (e.g. [newamerica.org/better-life-lab/subscribe](https://www.newamerica.org/better-life-lab/subscribe) / [newamerica.org/admin/pages/14/edit/#tab-promote](https://newamerica.org/admin/pages/14/edit/#tab-promote))

victoriachan:
Hi,

This isn't straightforward. The 'checked by default' option on the subscribe page CMS comes from the `SubscribePage` model `newsletter_subscriptions` streamfield field (https://github.com/newamericafoundation/newamerica-cms/blob/master/home/models.py#L433).

The React frontend here https://github.com/newamericafoundation/newamerica-cms/blob/master/newamericadotorg/assets/react/program-page/components/Subscribe.js (I think) is responsible for displaying the checkbox as checked or not. It probably just assumes the checkboxes all start out as unchecked. 

I can't find where in the API it is getting the subscription list from. We might need to forward this to a more experienced React developer.

I'll let Paul know to sort out the resourcing.

Regards,
Victoria

AnnaGrieve:
Need to assign a dev with React skills to look into this one

kaedroho:
It looks like we get a list of `subscriptions` passed in props. According to the code, each item in this list has a `title` field so I'm assuming this also contains `checked_by_default` as well.

If that's the case, we should be able to add some code into the constructor that initialises `state.subscriptions` with the subscriptions that have this field set to True.

I think this would take an hour.

AnnaGrieve:
To schedule in Dec support sprint

AnnaGrieve:
I have upped the dev estimate on this one slightly given that it is React based

kaedroho:
[...]

PR #1662, opened 2021-05-19
Subscription checkboxes

nmorduch:
- Checkbox styles for subscription page. They actually look like checkboxes, though I basically went with existing design.
- Only slight changes to radio buttons (see Publications page)

I would review this by commit. Some are messy but I did separate out the times I was just rearranging things.

Addresses #1396 but does not close it.


chigby (approved):
I'd say this change look great overall. I can't say I completely understand everything going on in some of the trickier Sass selectors/logic, but the end result is something I think we can all agree is a big benefit.

PR #1666, opened 2021-05-21
Wagtail 2.13

### LINKED ISSUE #1655: [2hr] Upgrade to Wagtaill 2.13

chigby:
Fixes #1655 

There actually did not seem to be any other things we needed to change based on this wagtail update. The autocomplete fields on survey pages looked fine to me, but I bumped the version of `wagtail-autocomplete` to the latest patch level just in case.


nmorduch (approved):
Looking good!

PR #1671, opened 2021-05-27
Collection pages

### LINKED ISSUE #1660: Add "Collection" publication page type

nmorduch:
Fixes #1660


collection/apps.py
```diff
@@ -0,0 +1,7 @@
+from __future__ import unicode_literals
+
+from django.apps import AppConfig
+
+
+class ThreadConfig(AppConfig):
```
chigby, line 6: IMO this doesn't need to mention `Thread` in the name

newamericadotorg/templates/components/card_list.html
```diff
@@ -27,7 +31,9 @@ <h6 class="card__text__excerpt margin-top-5 margin-bottom-0">
       <h6 class="inline margin-0">By:&nbsp;</h6>{% for author in post.post_author.all %}<h6 class="link inline margin-0"><a href="{{author.url}}">{{author.first_name}} {{author.last_name}}</a></h6>{% if forloop.counter == post.authors.all|length|add:"-1" and post.authors.all|length > 2 %}<h6 class="inline margin-0">, and&nbsp;</h6>{% elif forloop.counter == post.authors.all|length|add:"-1" %}<h6 class="inline margin-0">&nbsp;and&nbsp;</h6>{% elif forloop.counter != post.authors.all|length %}<h6 class="inline margin-0">,&nbsp;</h6>{% endif %}{% endfor %}
     </div>
     {% endif %}
-    <h6 class="card__text__program caption margin-top-5 margin-top-md-15 margin-bottom-0">{{post.parent_programs.all.0.title}} {{post|model_display_name}}</h6>
+    <h6 class="card__text__program caption margin-top-5 margin-top-md-15 margin-bottom-0">
+      {% firstof program post.parent_programs.all.0.title %} {% firstof page_type post|model_display_name %}
+    </h6>
```
chigby, line 36: I feel like in general we want to be somewhat wary of putting a lot of "variant names" into a template as eventually it can get a little unwieldy and hard to tell where a given piece of text is coming from. I don't actually think we've gone too far here, but it's something to think about.

An alternate approach is, rather than having a template that works no matter what "shape" the data you put into it, is to standardize the data prior to sending it to the template. This approach is also easier to write tests for.

But I think this is fine where it is.

chigby (dismissed):
I left a couple of comments but overall I think this is good. I'm marking this as approved but feel free to change things if the comments I did leave give you some kind of inspiration to do so :slightly_smiling_face:

collection/apps.py
nmorduch: lol this is from copying from the thread. thanks for catching :)

newamericadotorg/templates/components/card_list.html
nmorduch: That's a really fair point! This is definitely a case of doing things the quicker way that will hopefully not mess with existing stuff too much. I've mostly done it so that I leave the preexisting option, but add an override variable name, and generally didn't add too many additional options.

collection/templates/collection/collection.html
```diff
@@ -0,0 +1,88 @@
...
+		</div>
+		<div class="bg-beige" style="padding: 1px 0">
+			<div class="container margin-35 margin-lg-60">
+			{% if page.programs_list %}
+				<h2 class="margin-top-60">Programs</h2>
+				{{ page.programs_description|richtext }}
+				<div class="row gutter-10">
+					<div class="col-md-6 col-12">
+						{% for program in page.programs_list %}
+							<!-- {% cycle 'left' 'right' as column %} -->
```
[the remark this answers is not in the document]
nmorduch: this was a questionable (but quick to write) way to divide into two columns lol
It's in an html comment bc the cycle tag prints that text. This is definitely a silly way to write it but it didn't feel worth improving since these templates are probably short lived! (famous last words)

PR #1682, opened 2021-06-11
1681 google tag manager

### LINKED ISSUE #1681: Google Tag Manager

nmorduch:
Closes #1681


chigby (commented):
Retroactive review: I think this does exactly the right thing, and I'm confirming that I think the positioning of the `<script>` tag relative to the `<head>` tag is correct.

PR #1684, opened 2021-06-11
Improvements for subprograms with multiple parent programs

nmorduch:
- Route subprogram using in a RoutablePageMixin, so that the regex for finding a program url doesn't accidentally match every page with a path of the format `/foo/bar`. (And change all of the extra program routing to be done this way)
  - This also fixes a bug in which, because the subprogram was found by matching the subprogram slug without attention to the program slug, the view was sometimes rendering the subprogram that was intended just for redirecting, instead of the one with content and with things tagged to it. The example was [/education-policy/partnership-advance-youth-apprenticeship/subscribe](https://www.newamerica.org/education-policy/partnership-advance-youth-apprenticeship/subscribe/) which, on production and my local server, but not staging or dev, was rendering the CELNA subprogram that redirects to that page and had the same slug. Temporary fix was changing the slug of the CELNA one to include "to" to differentiate: [/center-education-labor/partnership-to-advance-youth-apprenticeship/](https://www.newamerica.org/center-education-labor/partnership-to-advance-youth-apprenticeship/)
- If a subprogram redirects to another subprogram, redirect it and its React-created subpages. (There was a commented out view for doing this, but it didn't work because of the aforementioned regex matching problem.)
- *Removed from PR*:
  - *Show up to two parent program names in a subprogram header. I don't think they tend to have more than 2, and more than 2 would require fussing with the CSS so it wouldn't use `position: absolute`. (2 can just squeak by, in my opinion.) Without doing this, there wasn't much value in a subprogram having multiple parents, because it never did anything with the one that wasn't in the url.*
  image
 
  - *Note that the breadcrumb stuff currently isn't amazing in general, though. e.g. [/education-policy/cyber-citizenship-initiative/events/teaching-cyber-citizenship-in-an-age-of-disinformation](https://www.newamerica.org/education-policy/cyber-citizenship-initiative/events/teaching-cyber-citizenship-in-an-age-of-disinformation/) points to EPP when direct parent is Cyber Citizenship Initiative, whose parents are EPP, **and** ISP*

### Testing
Go to the following **by going to the url directly** (not by clicking around)

- /education-policy/partnership-advance-youth-apprenticeship/subscribe/
  - [x] You should have content when you click "Our People"
- /international-security/cyber-citizenship-initiative/
  - [x] You should end up at /education-policy/cyber-citizenship-initiative/
- /international-security/cyber-citizenship-initiative/our-people/
  - [x] You should end up at /education-policy/cyber-citizenship-initiative/our-people/


nmorduch:
This also could stand to have tests added. I still need to get comfortable writing Python tests.

report/models.py
```diff
@@ -217,7 +217,7 @@ def print(self, request):
     @route(r'[a-zA-Z0-9_\.\-]*/$')
     def section(self, request):
         # Serve the whole report, subsection routing is handled by React
-        return self.serve(request)
+        return self.render(request)
```
chigby, line 220: I support this change :+1:

programs/models.py
```diff
@@ -398,6 +405,21 @@ def get_indexed_objects(cls):
...
+    @route(r'^(?:our-people|events|projects|publications|topics|about|subscribe)/$')
+    def redirect_to_subprogram(self, request):
+        redirect_page = getattr(self, 'redirect_page', None)
+        if redirect_page:
+            if isinstance(redirect_page.specific, Subprogram):
+                path = request.path.split('/')
+                redirect_path = redirect_page.url.split('/')
+                path[1] = redirect_path[1]
+                path[2] = redirect_path[2]
+                url = '/'.join(path)
```
chigby, line 419: It might be possible to simplify this code (or make it more readable, anyway) by changing the second `@route` decorator to capture the "sub-page" being requested (because of the `?:` in the regular expression it is not capturing it) and append that to the redirect page directly, i.e. something like

```python
@route(r'^$')
@route(r'^(our-people|events|projects|publications|topics|about|subscribe)/$')
def redirect_to_subprogram(self, request, section=''):
    redirect_page = getattr(self, 'redirect_page', None)
    if redirect_page:
        if isinstance(redirect_page.specific, Subprogram) and section:
            return redirect(redirect_page.url + section + '/')
        return HttpResponseRedirect(redirect_page.url)
    return self.render(request)
```

But it's up to you, I think, which one you prefer.

programs/tests.py
```diff
@@ -157,7 +195,22 @@ def test_subprogram_subpages(self):
...
+        response = c.get(self.article_project_page.url, follow=True)
+        self.assertEqual(response.redirect_chain, [(self.article.url, 302)])
+    def test_project_subprogram_redirect(self):
+        c = Client()
+        response = c.get(self.subprogram_project_page.url, follow=True)
+        self.assertEqual(response.redirect_chain, [(self.subprogram_page.url, 302)])
+    def test_project_subprogram_subpage_redirect(self):
+        c = Client()
+        response = c.get(self.subprogram_project_page.url+'our-people/', follow=True)
+        self.assertEqual(response.redirect_chain, [(self.subprogram_page.url+'our-people/', 302)])
```
chigby, line 213: These tests are great! Love 'em.

chigby (dismissed):
Approving this, since it works, but if my comments inspire you to change anything, feel free to do so and merge at will.

programs/models.py
nmorduch: Ooh yeah that _is_ nicer. I'd just been following the pattern of how they used to be done! (And I think I added the non-capturing at some point when I was playing around with the regex)

PR #1689, opened 2021-06-24
1674 Fix full-bleed image spacing

### LINKED ISSUE #1674: [1hr] Full-bleed image spacing is off when it has no image credit
e.g. https://www.newamerica.org/education-policy/reports/lessons-from-the-illinois-media-mentor-project/
image

nmorduch:
Now has margin below when it has no image credit
Also removed css that wasn't being used (bc the selector was wrong)

Compare one without a source: http://localhost:8000/education-policy/reports/lessons-from-the-illinois-media-mentor-project/
and one with a source (should be unchanged) http://localhost:8000/education-policy/reports/bets-job-strategy/

Fixes #1674

### DIFF
--- a/newamericadotorg/assets/react/report/components/Heading.scss
+++ b/newamericadotorg/assets/react/report/components/Heading.scss
@@ -76,13 +76,8 @@
     flex-direction: column;
   }
 
-  .card__image__background {
-    padding-bottom: 48.461538461%;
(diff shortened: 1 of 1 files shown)


chigby (approved):
This changed fixed the problem for me, locally, so, let's merge it! :sparkles:

PR #1712, opened 2021-07-23
Use correct URL for elasticsearch service

chigby:
Also disable security, otherwise authentication is required for search
requests.


nmorduch (approved):
Thank you :)

PR #1714, opened 2021-07-23
Require at least one author on report models

### LINKED ISSUE #1696: [1h📦] Report doesn't load if there are no authors
Making it truly optional for reports to have no authors seems like a bigger decision, so let's fix this by requiring on the backend that reports have at least one author

chigby:
Fixes #1696


nmorduch:
So in the process of testing this I realized that we do handle when other types of posts don't have authors, so I think we should do that here as well. PR coming shortly.

PR #1715, opened 2021-07-28
Add timestamp to webpack watch output

nmorduch:
And add npm script for more minimal output from webpack


chigby (approved):
Nice quality of life improvement here, and I also like the minimal watch command. I very well might start using that most of the time!

PR #1717, opened 2021-07-28
1696 authorless reports

### LINKED ISSUE #1696: [1h📦] Report doesn't load if there are no authors

nmorduch:
Closes #1696, supersedes #1714


chigby (approved):
Resolve the problem for me, I was able to open & see a report without any authors on it with this code, where it was broken on main.

PR #1718, opened 2021-07-28
Add slack notification for `workflow_submitted` signal

### LINKED ISSUE #1658: [2h] Slack integration for Wagtail workflow moderation submissions
We would like to have an integration so that when people submit a page for moderation, it pop up in a Slack channel (so we can discuss who's going to deal with it).

chigby:
Fixes #1658


nmorduch (commented):
Neither comment is blocking, but you could do either or both of them on this PR if you want

nmorduch:
I forgot one thought, which is that it might be nice to be able to separate it out by workflow, e.g. we might want to have a channel with people who approve EPP things that only has the EPP things. But that also seems like it would go in a different ticket.

chigby:
I made the following changes:
1. Moved the notification channel to an environment variable
2. Added notifications for approved, rejected, and cancelled.
3. Added the name of the user
4. Added a link to the history page

chigby (commented):
I think your additions are great, @nmorduch! This seems definitely ready to merge and start getting notifications. Once we've started using it, we can adjust as needed.

PR #1722, opened 2021-07-29
Exclude reports from the other pages search

chigby:
This pull request adds Reports to the list of page types excluded from the Other Pages search bucket. To be honest I'm not sure why this was not originally there, but it seems that it was not! It's there now, though.


nmorduch (approved):
Looks great!

Issue #1238, opened 2017-01-03
[2h👀📦] Set Security Headers


lomaxap:
Content Security Policy, HTTP Strict Transport Security, HTTP Public Key Pinning, X-Content-Type-Options, X-Frame-Options, X-XSS-Protection
Reference: [Mozilla's Observatory card](https://observatory.mozilla.org/analyze.html?host=newamerica.org)

nmorduch:
Does this still need changing? And how big a deal is this?

chigby:
Header | Situation
------ | ---------
Content Security Policy (CSP) | Step one is #1731 -- once that PR is merged and deployed we can start collecting more information about what content needs to be added to the CSP before we can activate it. Some time for data collection then ~1.5 hours.
HTTP Strict Transport Security | This feature can be [enabled in Django directly](https://docs.djangoproject.com/en/3.2/ref/middleware/#http-strict-transport-security) but requires some caution in rolling out, as improper SSL configuration could leave the domain unusable for some clients for a long time. Enabling this header is straightforward but I recommend we use a low `max-age` value to begin with (e.g. 1 hour), then increase it to the recommended 2 years when we're sure it's working. Also, to get maximum value from this we should submit the site to the [preload list](https://hstspreload.org/). Probably ~3hrs total, maybe more depending on how testing goes.
HTTP Public Key Pinning | I'm not sure we actually want to use this header at all. [The documentation](https://infosec.mozilla.org/guidelines/web_security#http-public-key-pinning) says its function is to instruct browser to use a specific SSL certificate authority or public key, which mitigates attacks by issuers of fraudulent certificates. It is not clear to me we are worried about this or what certificates we want to pin to address such worries. The documentation also says: "it is _not recommended_ for the majority websites to implement HPKP."
X-Content-Type-Options | Resolved in #1730 
X-Frame-Options | I think this is already being done. Our current site sends the header `X-Frame-Options: DENY`
X-XSS-Protection | Resolved in #1730

nmorduch:
Header | Situation
------ | ---------
Content Security Policy (CSP) | Requested changes
HTTP Strict Transport Security | Let's start with the short max age as you recommend, and with `SECURE_HSTS_INCLUDE_SUBDOMAINS` set to false because we have a bunch of subdomains and we can check them all later.
HTTP Public Key Pinning (HPKP) | Let's skip this like you suggest
X-Content-Type-Options | Resolved in #1730 
[...]

PR #1730, opened 2021-08-19
Add security headers for X-XSS-Protection and X-Content-Type-Options

chigby:
Refs #1238

This PR adds explicit settings for two security headers we should be using `X-XSS-Protection` and `X-Content-Type-Options`.


nmorduch (approved):
No problems I've found, and should be easy to switch back if we do run into problems!

PR #1731, opened 2021-08-19
Add initial CSP configuration

chigby:
Refs #1238 

This set of changes configures the CSP based on my initial clicking around the website locally, looking for problems.  Inevitably, there will be things I have missed (due to lack of exhaustive searching and differences between my environment and production).  Because of this, I have set the CSP middleware to send "report-only" headers, meaning the actual functioning of the site will not be affected yet.  It will report violations to a URI that we configure to receive that information, which we can then use to determine when we are ready to activate the CSP for real.

I've set the CSP_REPORT_URI to send it to Sentry.io's service for receiving these reports in the Heroku settings. 

This change should be safe to deploy as it should not change what resources the browser loads at all, only reports them if they would violate the set of directives I've initially added.


nmorduch (changes requested):
Let's do as Sentry recommends:
>We recommend setting this up to only run on a percentage of requests, as otherwise you may find that you've quickly exhausted your quota.

chigby:
Added a throttling of the CSP to 5%

nmorduch (approved):

PR #1735, opened 2021-08-26
Set the slug field for ProgramAboutHomePage to "about" pre-save

### LINKED ISSUE #1698: [1h] Automatically set slug for program about homepages to `about`
The links for the subpages don't come out right if the slug isn't `about`, but nobody knows that that's what it should be, and Wagtail automatically generates a slug based on the page title. Unless we entirely change how program about pages work, we should make sure the slug is `about` on save.

chigby:
Fixes #1698

### DIFF
--- a/home/tests.py
+++ b/home/tests.py
@@ -11,6 +11,7 @@ from article.models import AllArticlesHomePage, Article, ProgramArticlesPage
 from blog.models import AllBlogPostsHomePage
 from book.models import AllBooksHomePage
 from brief.models import AllBriefsHomePage
+from collection.models import CollectionsHomePage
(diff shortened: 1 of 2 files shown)


home/tests.py
```diff
@@ -11,6 +11,7 @@
 from blog.models import AllBlogPostsHomePage
 from book.models import AllBooksHomePage
 from brief.models import AllBriefsHomePage
+from collection.models import CollectionsHomePage
```
nmorduch, line 14: 👍🏻

nmorduch (approved):
Very nice! I wish there were an easy way to tell people this were going to happen, but tbh they're mostly just not going to notice that anything happened.

PR #1736, opened 2021-08-26
Remove `image` feature from acknowledgements field

### LINKED ISSUE #1703: [1.5h] Rich text field for report acknowledgements allows image but then doesn't display them
Possible solutions include:

- Don't do image embedding inside this rich text field
- Don't do image embedding inside any rich text fields (prob good tbh but check on implications. Maybe just hide button?)

chigby:
Fixes #1703


nmorduch (approved):
We'll have to remember to update this when we change what the default features are in general, but I don't think that's a problem! Also worth noting that it appears that if there had been an image in that field, it will disappear now.

PR #1737, opened 2021-08-26
Move fellowship year calculation from model to form widget

### LINKED ISSUE #1471: [1hr] Fellowship year breaks migrations yearly
https://github.com/newamericafoundation/newamerica-cms/blob/830f28d3d45a3d8fe4eb4473cc1aa98a3c47b202/person/models.py#L35

So, every year this starts expecting there to be one more option.

Karl says:
>I think we could either change it so the choices are set on the form instead of the model. Or we could remove the choices and add validation instead
>(latter would probably be easier but the input would change)

chigby:
Fixes #1471


nmorduch (approved):
An elegant solution!

PR #1740, opened 2021-09-23
Enable 1-hour HTTP Strict Transport Security headers 

### LINKED ISSUE #1744: [3h] HTTP Strict Transport Security
@chigby in a [comment](https://github.com/newamericafoundation/newamerica-cms/issues/1238#issuecomment-902913687) on #1238 
>This feature can be enabled in Django directly  but requires some caution in rolling out, as improper SSL configuration  could leave the domain unusable for some clients for a long time.  Enabling this header is straightforward but I recommend we use a low max-age  value to begin with (e.g. 1 hour), then increase it to the recommended 2  years when we're sure it's working. Also, to get maximum value from this we should submit the site to the preload list. Probably ~3hrs total, maybe more depending on how testing goes.

@nmorduch [replying](https://github.com/newamericafoundation/newamerica-cms/issues/1238#issuecomment-925858881)
>Let's start with the short max age as you recommend, and with `SECURE_HSTS_INCLUDE_SUBDOMAINS` set to false because we have a bunch of subdomains and we can check them all later

chigby:
In accordance with Django's guidelines, I've set the `max-age` to a short amount of time in order to ensure everything is working without locking anyone out of the site. If everything goes all right we can increase it.

Refs #1238 
Fixes #1744


nmorduch:
I want to test this on the na-develop when we're done using that for elasticsearch

nmorduch (dismissed):

chigby:
The header seems like it's working on https://na-develop.herokuapp.com/ with max age = 1 hour

nmorduch (approved):

PR #1743, opened 2021-09-27
Add partner logo on single page reports

nmorduch:
Also make the margin class one that's real.


chigby (approved):
Looks great! Worked for me locally.

PR #1745, opened 2021-09-29
Update react-slick

nmorduch:
The earlier version was unhappy about the version of react we're using


chigby (approved):
Update looks good to me!

PR #1753, opened 2021-09-30
Additional csp updates

chigby:
This pull request:

1. Fixes a couple of typos
2. Adds a few more sources. I'm now no longer getting any CSP errors on the home page except for inline scripts, which I think should be converted into static JS files. That can be another PR, though.

### DIFF
--- a/newamericadotorg/settings/base.py
+++ b/newamericadotorg/settings/base.py
@@ -286,6 +286,7 @@ CSP_SCRIPT_SRC = (
     "'unsafe-inline'",  # Needed for Wagtail admin pages
     'https://www.googletagmanager.com',
     'static.ads-twitter.com',
+    'https://analytics.twitter.com',
     'https://www.google.com/recaptcha/',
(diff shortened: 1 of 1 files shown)


chigby:
This should be ready for review. I think I was wrong about having to move our inline scripts into their own files, as wagtail actually requires inline scripts for the admin to function, so we have to have `unsafe-inline` no matter what we do.

chigby:
Not sure what's going on with the CI here exactly, but I feel like this is ready to move forward.

newamericadotorg/settings/base.py
```diff
@@ -332,11 +326,13 @@
     'https://www.google-analytics.com',
     'https://stats.g.doubleclick.net',
     'https://sumo.com',
+    'http://sumo.com',
```
nmorduch, line 329: I'm annoyed we have an http one but it's probably out of our hands 🤷🏻‍♀️

PR #1759, opened 2022-06-15
Add 2022 active funding table

nmorduch:
Waiting on data for table


nmorduch:
You can ignore the Travis CI! We're only doing the Github one

PR #1765, opened 2023-04-04
Add sentry environment

nmorduch:
Looking here: https://docs.sentry.io/platforms/python/guides/celery/configuration/environments/

I think we just didn't have environments set up in the code, even though we already have the variable in Heroku.


nmorduch:
It's working on na-develop!

chigby (approved):
Looks great :)

Issue #1746, opened 2021-09-29
[1h] Add sites to CSP configuration


nmorduch:
Follow up to #1731

nmorduch:
Is there a reason some of the sources don't list the protocol at all? Ideally they'd all be https, but even if we have to add some http in there, I'd rather not leave them unspecified if we don't have to.

PR #1766, opened 2023-04-07
More CSP Updates

chigby:
Refs #1746

1. Adds "https" wherever there was a protocol missing.
2. Adds additional allowed domains based on sentry alerts
3. Adds "*.newamerica.org" to more directives (this is necessary)

### DIFF
--- a/newamericadotorg/settings/base.py
+++ b/newamericadotorg/settings/base.py
@@ -285,14 +285,14 @@ CSP_SCRIPT_SRC = (
     "'unsafe-eval'",  # Needed for funding disclosures JS
     "'unsafe-inline'",  # Needed for Wagtail admin pages
     'https://www.googletagmanager.com',
-    'static.ads-twitter.com',
+    'https://static.ads-twitter.com',
     'https://analytics.twitter.com',
     'https://www.google.com/recaptcha/',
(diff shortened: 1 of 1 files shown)


newamericadotorg/settings/base.py
```diff
@@ -284,51 +284,57 @@
...
+    'https://datawrapper.dwcdn.net',
     'https://www.google.com',
     # Embeds
     'https://www.youtube.com',  # YouTube
     'https://w.soundcloud.com',
     'https://airtable.com',
     'https://art19.com',
+    'https://renderer.apester.com',
 )
 CSP_OBJECT_SRC = ("'self'")
```
nmorduch, line 333: Should we also add `'https://*.newamerica.org'` here? Is there a good way to have each one also include the default source, like make each list include that list? Or is that silly

chigby: Hmm, if `object-src` is just `'self'` right now, it's maybe the best thing to do is remove it entirely, and the browser will automatically fall back to `default-src`

nmorduch: Good point. Let's do that!

chigby: Done!

nmorduch (approved):

PR #1767, opened 2023-04-14
Csp updates

nmorduch:
added then sorted


chigby (approved):
Everything here looks good, and I think sorting them is helpful

PR #1768, opened 2023-04-18
CSP updates

nmorduch:


newamericadotorg/settings/base.py
```diff
@@ -334,44 +338,51 @@
     'https://www.google.com',
     'https://www.googletagmanager.com',
     # Embeds
-    'http://www.ustream.tv',
```
nmorduch, line 337: This doesn't actually seem to exist anymore, so no value in including this one that has http

newamericadotorg/settings/base.py
```diff
@@ -334,44 +338,51 @@
...
     'https://syndication.twitter.com',
     'https://w.soundcloud.com',
     'https://www.facebook.com',
+    'https://www.flickr.com',
     'https://www.instagram.com',
     'https://www.youtube.com',
 )
 CSP_CONNECT_SRC = (
     "'self'",
-    'http://sumo.com',
```
nmorduch, line 362: Let's leave it off and see how it goes, since this is our only other http one

chigby (approved):
:sparkles: Looks good!

PR #1771, opened 2023-04-24
Add additional sources to CSP

nmorduch:


chigby (approved):
:sparkles: More domains :sparkles:

PR #1776, opened 2023-05-19
Alt text fields

### LINKED ISSUE #1775: Add alt text fields
See https://docs.wagtail.org/en/stable/advanced_topics/accessibility_considerations.html#alt-text-for-images

Story images also need alt text, and we should check for other places that have images

chigby:
Fixes #1775

This PR adds alt text fields to:

1. Page models that have an image somewhere
2. Blocks that have an image somewhere

It does not update the front end or any of the templates to render the contents of these fields.

One place I wasn't sure how to address was the In Depth Profile page, which has the option to include an image its StreamField body. This should probably be changed to a different block type that includes an option for alt text, but that might require some additional front-end updates (and/or a data migration on the back-end to use a new block type), and I'm not sure if that's part of the scope of this issue.


nmorduch:
I've added some of the FE but not all yet. Would you mind adding the alt text fields to the serializers as needed? That would be very helpful to me!

nmorduch (approved):

PR #1778, opened 2023-05-25
Navigation accessibility

nmorduch:


nmorduch:
reference for skip link btw https://webaim.org/techniques/skipnav/

chigby (approved):
Many big improvements here. The mobile submenus don't seem 100% there, it takes a few tabs to get to the newly opened items and the back button seems inaccessible, but lots of other things are doing very well.

PR #1799, opened 2023-07-21
Person ordering flexibility

### LINKED ISSUE #1795: More flexible sort order for people
- [ ] Sort order should be a number field rather than a dropdown
- [ ] The person program relationship should also have a sort order, and the one that's not about a person program relationship should not apply in a program situation
- [ ] Also add help text clarifying some of this

chigby:
Fixes #1795

1. There are a number of migrations here, but they're necessary for preserving the existing `sort_priority` values by copying them into the new `sort_order` field (which is now an integer field rather than a list of preset choices).
2. The new field on program/subprogram relationships is also called `sort_order` and it's working pretty well for me, locally. For example, on some subprograms I picked at random:

image

image


newamericadotorg/api/author/tests.py
```diff
@@ -103,3 +106,152 @@ def test_get_fellows(self):
...
+    def setUpTestData(cls):
+        home_page = PostFactory.create_homepage()
+
+        program1 = PostFactory.create_program(home_page=home_page)
+        program2 = PostFactory.create_program(home_page=home_page)
+        subprogram1 = PostFactory.create_subprogram(program=program1)
+
+        person1 = PostFactory.create_person(
+            {
+                "sort_order": 2,
```
nmorduch, line 122: I think you also need to test with people not having a sort order! Same for the person program relationship

nmorduch (changes requested):
See comment re testing!

newamericadotorg/api/author/tests.py
nmorduch: To be clear, people with no sort order set go last!

chigby:
Added the tests for absent sort order

nmorduch (approved):

PR #1801, opened 2023-07-28
Add option for people to be ASU fellows

nmorduch:


chigby (dismissed):
Looks good and works for me locally, just needs a clean up of the conflict with main.

chigby (approved):

PR #1808, opened 2023-09-08
Subscription Models, Part 1

chigby:
This pull request spins off the preliminary database setup from https://github.com/newamericafoundation/newamerica-cms/pull/1792 into its own branch. The purpose of this is to create new snippet models and page types that can be populated with the `make_subscribe_pages` command but won't affect the functionality of the site. This way we can set up the data beforehand and deploy the site changes in another release at a later date.

Order of operations:
- [ ] deploy #1808 (this branch)
- [ ] run `make_subscribe_pages`
- [ ] make subscribe page for the thread
- [ ] deploy #1792 
- [ ] run `clear_cache`
- [ ] update which are checked by default
- [ ] run `clear_cache` again


home/models.py
```diff
@@ -485,7 +485,11 @@ class SubscribePage(OrgSimplePage):
     """
     Subscribe Page at the organization level
     """
-    parent_page_types = ['home.HomePage']
+    parent_page_types = [
```
nmorduch, line 488: Also needs to be able to go under the Thread homepage

chigby: Added this!

nmorduch (changes requested):
Working great! One last change: Let's have all the new segments on the subscribe pages checked by default.

nmorduch (approved):
Perfect. Will wait to merge until the other is ready.

PR #1809, opened 2023-09-15
Remove weekly subscribe box and lead article

nmorduch:
Remove subscribe box because it doesn't make sense to have anymore, and make the lead a regular article so that there are the right number of articles for a 3 col grid


chigby (approved):
The weekly page looks fine to me after this change, so I'm approving :+1:

PR #1810, opened 2023-09-15
Remove fellowship_year field, which wasn't used

nmorduch:


chigby (approved):
Looks great, I concur that this field is not being used anywhere right now.

PR #1815, opened 2023-10-13
Elasticsearch 7

chigby:
Based off of `subscribe-page-models` branch. Should not be merged until after #1813

This PR updates us to elastic search (ES) version 7. This includes our custom search query code, the ES client library, and the dev environment container. It does not include any changes to the heroku add-ons, which will have to be made in conjunction with the deployment. 

For this code to work, `manage.py update_index` _must_ be run in whatever environment you are trying to use.


chigby:
Rebased this branch on the wagtail 5.2 branch, it should only be merged after that one.

chigby:
Having problems when deploying this branch on the `na-develop` app on Heroku after re-provisioning ES 7 there on searchbox. The `update_index` command succeeds, wagtail admin search seems to work, but our search API queries (events, people, etc.) all seem to fail with [this error, ultimately](https://new-america.sentry.io/issues/4815586467/?project=2985675&query=is%3Aunresolved&referrer=issue-stream&stream_index=2). Tried reproducing this locally but was not successful. 

I thought this might be because of a mismatch between the version of ES on searchbox (7.10.2) and the version I had installed locally and on CI (7.17), but I downgraded the local version to 7.10.2 (and also the version of the python ES client library), and it still worked locally. Not sure if this is super-worth pursuing further.

PR #1819, opened 2023-11-03
Add management command to tag posts into a given program

chigby:
Management command that takes two ID arguments (and an optional `commit` flag) and:

1. Looks for a page with the `to` ID and collects all pages with this page as an ancestor.  This collection is the set of "child" posts.
2. Looks for a program, subprogram, or issue/topic with the `from` slug.
3. If both are found, two actions are performed:
    1. The `save()` method is called on all the child posts of the `to` Program.
    2. All relationships between the `from` Program and _any_ Post or Person are removed, regardless of where that Post or Person is located in the page hierarchy. These relationships are replaced by relationships to the `to` Program.

By default, only provides counts but does not alter the database. Only make changes if `--commit` is given.  Otherwise, display information about what would be done.

Usage examples:

```
$ docker compose exec web ./manage.py retag_program_posts --to 1616 --from 1610
```

```
$ docker compose exec web ./manage.py retag_program_posts --to 1616 --from 1610 --commit
```

### DIFF
--- a/home/management/commands/retag_program_posts.py
+++ b/home/management/commands/retag_program_posts.py
@@ -9,23 +9,28 @@ from home.models import (
     PostTopicRelationship,
 )
 from issue.models import IssueOrTopic
+from person.models import (
+    PersonProgramRelationship,
+    PersonSubprogramRelationship,
+    PersonTopicRelationship,
+)
 from programs.models import Program, Subprogram
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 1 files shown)


nmorduch (changes requested):
- [ ] use id instead of slug bc slug can (and will) not be unique
- [ ] change to: under page a, untag all page b from all Posts, then run save on all Posts

chigby:
Updated the management command and the description of it in the description of this PR.

nmorduch (approved):

nmorduch (approved):

PR #1825, opened 2023-12-15
Fold locals into regular program list

nmorduch:
This requires setting NA CA and NA Chicago not to be marked as being location based.

### DIFF
--- a/home/templatetags/utilities.py
+++ b/home/templatetags/utilities.py
@@ -287,7 +287,7 @@ def split_program_list(start_list, max_size=6):
 
 
 @register.simple_tag()
-def group_programs(programs, cols=3):
+def group_programs(programs, cols=4):
     length = len(programs)
     l = length / float(cols)
     max_items = math.ceil(l)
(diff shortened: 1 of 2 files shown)


home/templatetags/utilities.py
```diff
@@ -287,7 +287,7 @@ def split_program_list(start_list, max_size=6):
 
 
 @register.simple_tag()
-def group_programs(programs, cols=3):
```
chigby, line 290: I think we don't need to change the default argument here? We could call `{% group_programs program_data 4 as program_groups %}` in the template. Though it also seems like that's the only place on the site we're using the default argument so maybe it doesn't matter.

chigby (approved):
I think this is good change, and it works as described. I've left a comment that you can inspire you to make modifications or leave as is.

home/templatetags/utilities.py
nmorduch: I'm going to leave it but that's a good point!

PR #1829, opened 2024-02-02
fix link color in pdf

nmorduch:


chigby (approved):
Color change looks good!

PR #1833, opened 2024-02-09
Add auto-generated ID value to heading tags in reports

chigby:
Work in progress. The code to create the IDs is there, but I'm not sure where else it ought to be applied.

### DIFF
--- a/home/templatetags/anchorize_headings.py
+++ b/home/templatetags/anchorize_headings.py
@@ -1,0 +1,36 @@
+from collections import Counter
+
+from django import template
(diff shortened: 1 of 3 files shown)


nmorduch:
I think [templates/components/post_body.html](https://github.com/newamericafoundation/newamerica-cms/blob/main/newamericadotorg/templates/components/post_body.html) would be a good place to add it too! And then we can call it ready to go.

chigby:
Added the same `id` treatment to the `post_body` template

PR #1841, opened 2024-03-08
Feature page ordering

### LINKED ISSUE #1840: Reverse pages ordering
When a program/subprogram has a lot of featured pages, adding one to the top requires adding one at the bottom. To fix this, reverse them both in the frontend and the backend (which means they display in the same order as before)

- [ ] Reverse the order in which featured pages are displayed on Programs/Subprograms, which means that newly added ones are by default featured first
- [ ] Reverse the saved order of the featured pages, so that the content on the frontend will remain unchanged
- [ ] Update the note `First page becomes lead story` to `CHANGED MARCH 2023: Pages are shown in reverse order. The last page in this list is shown as the lead story.`

chigby:
Fixes #1840

This pull request makes adjustments needed to display featured pages on Program home pages in descending order by the `sort_order` field. This field is the one that is determined page a page's up-and-down position on an `InlinePanel`. The front end program page components make requests to two different API endpoints to display these featured pages: the program details view, and the program featured page list. Both of these were hard-coding the ordering of featured pages to be by `sort_order` ascending. I've changed it so that the front end can now specify whether it wants the ordering to be ascending or descending. The ordering cannot easily be done on the front end entirely, as the featured pages are loaded in in "pages" of seven at a time.

So I've updated the front end to ask for descending order in all the places I could find.

I've also added a data migration that inverts the `sort_order` field for all featured pages on all Programs (except for a few that were unordered to start with--not sure how that happened. Some programs' featured pages have all `NULL` in the `sort_order` column. These have been unchanged). This will preserve the current presentation of featured pages.

Also, the help text has been updated to reflect these changes:
image


nmorduch:
Things to test:
- [ ] featured pages are in the correct order, taking into account pagination
- [ ] in the admin featured pages are in reverse order from their previous order
- [ ] the note makes sense

nmorduch (changes requested):
The sort order of `/api/program/{pk}/featured/` seems to be backwards of what it should be, but otherwise looks good!

nmorduch (approved):
Works great now! Thank you!

nmorduch (approved):

PR #1878, opened 2024-10-16
Support external links for homepage lead stories (Part 2)

SharmaineLim:
**Dependent on:** #1877 | In draft only to prevent merging & deployment before its time

### When applied, this PR will...

remove the legacy lead story fields.

#### Why 2 pull requests?

If we make this change as 1 pull request, there will be a period of time (between the deployment and an editor updating the homepage) where the homepage will not have any lead stories at all.

Separating this into 2 pull requests means that we can first introduce the new field with external link support, then the editor can make the changes needed, and then remove the old fields. This gives a seamless transition for site viewers.


jams2 (approved):
The changes make sense to me @SharmaineLim

PR #1880, opened 2025-02-05
Add Full Page Embed page type

tomusher:
This adds a Full Page Embed page type which allows embedding a URL as an iframe, scaled to 100%.

The page is not indexed in search.


stephen-darling (approved):
looks good to me!

PR #1881, opened 2025-02-05
Bump celery version

tomusher:
Trying to deploy https://github.com/newamericafoundation/newamerica-cms/pull/1880 to Heroku failed with an error

```       ERROR: No matching distribution found for celery==4.4.7
 !     Error: One of your dependencies contains broken metadata.
 !     
 !     Newer versions of pip reject packages that use invalid versions
 !     in their metadata (such as Celery older than v5.2.1).
 !     
 !     Try upgrading to a newer version of the affected package.
 !     
 !     For more help, see:
 !     https://devcenter.heroku.com/changelog-items/3073
 !     Error: Unable to install dependencies using pip.
 !     
 !     See the log output above for more information.
```

This is because the Heroku builder uses an updated version of `pip` which is more strict about broken metadata in older packages.

While we could find an alternative approach to building the app (with an older version of `pip`), having reviewed the breaking changes in Celery releases up to latest, I can't see anything (other than the change to the `Procfile` addressed in this PR) that would cause issues.

Bumping the version fixes the issue and should enable us to deploy successfully. We can test Celery behaviour on staging before promoting to production.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,3 +1,3 @@
 release: python manage.py migrate
 web: gunicorn newamericadotorg.wsgi --log-file -
-worker: celery worker --app=newamericadotorg.celery.app --loglevel=INFO
+worker: celery --app=newamericadotorg.celery.app --loglevel=INFO worker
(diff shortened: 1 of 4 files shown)


stephen-darling (approved):
approved!

PR #1883, opened 2025-02-13
Embed page layout tweaks

tomusher:
This tweaks the embed page layout so it's full-width even when the header is shown.


stephen-darling (commented):
approved!

stephen-darling (approved):
approved!

PR #1886, opened 2025-03-07
Remove Shorthand when navigating between report sections

tomusher:
This change removes the Shorthand visualisation when navigating between report sections, which allows the Shorthand embed code to re-render the embed when the page is loaded again.


stephen-darling (approved):
looks good!

